PR #103, opened 2023-02-01
Artifact Store Integration

turingcompl33t:
This PR integrates the MLTE artifact store with the primary `mlte` repository:

- The source for the server is integrated with the primary codebase here, under the `store` submodule
- The filesystem backend implementations are unified. That is, when we run a local artifact store, or when we run an HTTP-based artifact store that uses the local filesystem as its storage location, we are now running the same implementation under the hood.
- The artifact store is exposed via a commandline interface:

```bash
mlte store --backend-uri fs://path/to/store
```

The artifact store server implementation is significantly overhauled from the original instantiation. The biggest change is a removal of the reliance on the dependency injection system from `fastapi`. This is a great feature, but it made it impossible to make the server work with both unit tests AND a standalone command line interface. I switched it over to a more "traditional" implementation based on a piece of shared global state.

The server implementation has also been refactored to make it easy to integrate new backends. I plan to start this with both a local and a remote `sqlite` backend that will dramatically increase the robustness of `mlte`.

Furthermore, it packages the artifact store as a Docker container image that we can publish to make the process of standing up an external artifact store easier.

```bash
docker run --rm -p 8080:8080 -v /host/path/to/store:/mnts/store mlte-store
```

By default, this executes with a filesytem backend within the container.

The things that this PR leaves incomplete are:

- We still need a refactor of the interaction between the low-level artifact store API and the rest of the package. How are artifacts from the artifact store exposed in application code? We have a mechanism in place now, but I do not think it is the final answer.
- We need to publish the `mlte-store` image on Dockerhub. We should update CI to perform this action automatically on a new merge to master.
- We need to bump the version and publish a new package.

I am leaving each of these to a separate PR because this one is already sufficiently large and complex.


turingcompl33t:
@sebastian-echeverria I am planning to merge this PR soon (finally) because it has been outstanding for so long. I finally got around to making the tweaks to complete this integration this weekend. The points in the PR comment above highlight what remains incomplete, but I plan to address these in separate PRs. I hope that this does not impact your work too much. The internals of the artifact store are unchanged; it is only the interface that has been updated substantially.

krmaffey (approved):
Great work Kyle. LGTM!

PR #105, opened 2023-02-07
Demo fixes

sebastian-echeverria:
Minor fixes mainly to demo code:
 - Adding code to automatically create folders used during the demo in case they don't already exist in the code where the demo is being run.
 - Added a requirements file with the libraries needed to un the the demo sample notebook.
 - Updated once dependency from the main requirements_dev file, since it was not working properly with the rest and it was giving an out of date error (importlib-metadata).

### DIFF
--- a/demo/evidence.ipynb
+++ b/demo/evidence.ipynb
@@ -127,6 +127,7 @@
     "import mlte\n",
     "\n",
     "store_path = os.path.join(os.getcwd(), \"store\")\n",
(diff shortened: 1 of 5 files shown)


turingcompl33t (approved):
Thanks for these updates @sebastian-echeverria. CI checks are failing, but I have addressed the CI environment issues in #112. I am going to merge this and then reconcile the two after the fact.

PR #111, opened 2023-02-15
Feature/external measurements

sebastian-echeverria:
To review process measurement improvements in a discussion. Note that this include PR #105, due to how this was tested, so that PR should be merged first before doing anything about this, and updating this one to not have that explicitly.


src/mlte/_private/job.py
```diff
@@ -0,0 +1,56 @@
+import sys
```
turingcompl33t, line 1: Can we add a brief file-level comment that describes the content of this file?

src/mlte/_private/job.py
```diff
@@ -0,0 +1,56 @@
...
+    :type script: str
+
+    :param arguments: A list of string arguments for the process.
+    :type arguments: List[str[]
+
+    :return: Process id of the process that was created.
+    :rtype: int    
+    """
+    python_executable = _get_interpreter_path()
+    new_arguments = [script_path]
```
turingcompl33t, line 32: Can use

```python
new_arguments = [script_path, *arguments]
```

to avoid initialize-then-modify antipattern.

src/mlte/_private/job.py
```diff
@@ -0,0 +1,56 @@
...
+    :param script: The full path to an executable process to run.
+    :type script: str
+
+    :param arguments: A list of string arguments for the process.
+    :type arguments: List[str[]
+
+    :return: Process id of the process that was created.
+    :rtype: int    
+    """
+    command = [executable]
```
turingcompl33t, line 50: Same as above. IMT can be avoided through iterable unpacking.

src/mlte/measurement/cpu/local_process_cpu_utilization.py
```diff
@@ -228,3 +228,5 @@ def _get_cpu_usage(pid: int) -> float:
         return -1.0
     except ValueError:
         return -1.0
+    except FileNotFoundError as e:
+        raise RuntimeError("External program needed to get CPU usage was not found: " + str(e))
```
turingcompl33t, line 232: To be consistent with string formatting across the codebase, can we refactor this to use f-strings:

```python
raise RuntimeError("External program needed to get CPU usage was not found: {e}")
```

src/mlte/measurement/memory/local_process_memory_consumption.py
```diff
@@ -239,3 +239,5 @@ def _get_memory_usage(pid: int) -> int:
         return int(used.decode("utf-8").strip()[:-1])
     except ValueError:
         return 0
+    except FileNotFoundError as e:
+        raise RuntimeError("External program needed to get memory usage was not found: " + str(e))
```
turingcompl33t, line 243: Same as previous comment. Refactor to f-string.

src/mlte/measurement/process_measurement.py
```diff
@@ -0,0 +1,94 @@
+"""
+Base class for measurement of external processes asynchronally.
```
turingcompl33t, line 2: "asynchronously"

src/mlte/measurement/process_measurement.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def __init__(self, instance: ProcessMeasurement, identifier: str):
+        """
+        Initialize a new ProcessMeasurement measurement.
+
+        :param identifier: A unique identifier for the measurement
+        :type identifier: str
+        """
+        super().__init__(instance, identifier)        
+        if is_windows():
```
turingcompl33t, line 47: Do you not intend to support `ProcessMeasurement` on Windows? The checks from `LocalProcessCPUUtilization` and `LocalProcessMemoryConsumption` that `raise` on Windows are there because the particular way that we measure these metrics is specific to *nix platforms, but spawning a Python subprocess should work fine on both *nix and Windows (thanks to the abstractions Python's stdlib provides).

src/mlte/measurement/process_measurement.py
```diff
@@ -0,0 +1,94 @@
...
+        self.thread.start()
+ 
+    def _run_call(self, pid, *args, **kwargs):
+        """
+        Runs the internall __call__ method that should implement the measurement, and stores its results when it finishes.
+        """
+        try:
+            self.result = self.__call__(pid, *args, **kwargs)
+        except Exception as e:
+            self.error = "Could not evaluate process: " + str(e)
```
turingcompl33t, line 77: Refactor to f-string.

src/mlte/measurement/process_measurement.py
```diff
@@ -0,0 +1,94 @@
...
+    def wait_for_result(self) -> Result:
+        """
+        Needed to get the results of a measurement executed in parallel using evaluate_async. Waits for the thread to finish.
+
+        :return: The result of measurement execution, with semantics
+        :rtype: Result
+        """
+        # Wait for thread to finish, and return results once it is done.
+        while self.thread.is_alive():
+            time.sleep(1)
```
turingcompl33t, line 88: Can we make the poll interval an optional argument?

src/mlte/measurement/process_measurement.py
```diff
@@ -0,0 +1,94 @@
+"""
+Base class for measurement of external processes asynchronally.
+"""
+
```
turingcompl33t, line 4: Can we add unit tests (standalone, and within the measurements that subclass `ProcessMeasurement`) to verify this functionality?

turingcompl33t: We don't have any code-coverage rules in place so far, but this is something I would like to add in the future, so keeping up with it now reduces the burden.

src/mlte/measurement/process_measurement.py
```diff
@@ -0,0 +1,94 @@
...
+    def _run_call(self, pid, *args, **kwargs):
+        """
+        Runs the internall __call__ method that should implement the measurement, and stores its results when it finishes.
+        """
+        try:
+            self.result = self.__call__(pid, *args, **kwargs)
+        except Exception as e:
+            self.error = "Could not evaluate process: " + str(e)
+
+    def wait_for_result(self) -> Result:
```
turingcompl33t, line 79: We should consider adding some mechanism to wait for all asynchronous results in a collection. This can be a separate feature / pull-request through.

turingcompl33t (changes requested):
Requesting a couple minor changes.

src/mlte/_private/job.py
sebastian-echeverria: Sure! We can restructure or move this somewhere else if needed, I guessed this could be a good location.

src/mlte/_private/job.py
sebastian-echeverria: Sure. Your linter seems to be more strict that what I am using. Could you remind me what you are using to check, so I can incorporate this into my flow, and detect these things earlier?

src/mlte/_private/job.py
sebastian-echeverria: Yep.

src/mlte/measurement/cpu/local_process_cpu_utilization.py
sebastian-echeverria: No problem

src/mlte/measurement/memory/local_process_memory_consumption.py
sebastian-echeverria: Ok

src/mlte/measurement/process_measurement.py
sebastian-echeverria: I gues I was sleepy when writing that.

src/mlte/measurement/process_measurement.py
sebastian-echeverria: Ok

src/mlte/measurement/process_measurement.py
sebastian-echeverria: Sure

src/mlte/measurement/process_measurement.py
sebastian-echeverria: I agree that we should do that (it could even be with Python's futures and executor), but I also agree this can be a separate feature since it could need more work or change a bit the structure.

src/mlte/measurement/process_measurement.py
sebastian-echeverria: Ok, will do

src/mlte/measurement/process_measurement.py
sebastian-echeverria: That makes sense, I just copied that there before I fully implemented it, and didn't really review it it made sense later.

src/mlte/_private/job.py
sebastian-echeverria: No worries, found them.

turingcompl33t (approved):
Thanks @sebastian-echeverria! LGTM.

PR #112, opened 2023-02-16
Development Dependencies Cleanup

### LINKED ISSUE #84: Update Development Environment Setup
We should take further steps to ensure a consistent development environment for both local users and in CI. Right now, we hack things a bit by differentiating CI from the local development environment setup (e.g. [here](https://github.com/mlte-team/mlte/blob/c41ccd364ca502e1d65f696a1292048d658db282/.github/workflows/tests.yaml#L24)). We should unify these approaches,  and ensure they work across Python versions as expected.

### LINKED ISSUE #100: Update Pipeline Configuration
The pipeline configuration is ad-hoc. We should refactor this setup such that the environment is consistent with the development environment for `mlte`, and furthermore such that the configuration for tools is consistent between local and CI invocations.

turingcompl33t:
This PR accomplishes the following:
- Specifies pinned versions for development dependencies that are specific to the version of Python in use. A separate `requirements.txt` file is generated for each supported minor version.
- Ensures consistent behavior between environments in CI and in local QA and testing.

Resolves #84.
Resolves #100.


krmaffey (approved):
LGTM!

PR #115, opened 2023-03-01
Feature/generic measurements

sebastian-echeverria:
* Added ExternalMeasurement measurement type, used to wrap external measurements by ensuring the return type is a Result type passed in the constructor, and allowing the use of evaluate() in a similar way to other measurements, only just wrapping the external function call.
 * Updated demo and docs to show use of ExternalMeasurement.
 * Very minor cleanup to evidence demo to run train_model() in a more efficient location (and more easily avoid re-calling it when not needed).

### DIFF
--- a/src/mlte/measurement/__init__.py
+++ b/src/mlte/measurement/__init__.py
@@ -9,5 +9,5 @@ __all__ = [
     "MeasurementMetadata",
     "Identifier",
     "ProcessMeasurement",
-    "ExternalMeasurement"
+    "ExternalMeasurement",
 ]
--- a/src/mlte/measurement/__init__.py
+++ b/src/mlte/measurement/__init__.py
@@ -2,10 +2,12 @@ from .measurement import Measurement
 from .measurement_metadata import MeasurementMetadata
 from .identifier import Identifier
 from .process_measurement import ProcessMeasurement
+from .external_measurement import ExternalMeasurement
 
 __all__ = [
     "Measurement",
     "MeasurementMetadata",
     "Identifier",
     "ProcessMeasurement",
+    "ExternalMeasurement"
 ]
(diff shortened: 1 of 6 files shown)


src/mlte/measurement/__init__.py
```diff
@@ -2,10 +2,12 @@
 from .measurement_metadata import MeasurementMetadata
 from .identifier import Identifier
 from .process_measurement import ProcessMeasurement
+from .external_measurement import ExternalMeasurement
 
 __all__ = [
     "Measurement",
     "MeasurementMetadata",
```
turingcompl33t, line 9: We should now be able to remove `MeasurementMetadata` and `Identifier` from the `measurement` module-level exports. I think that since we now have your mechanism for integrating external measurements in a more "familiar" way, we should just hide these details from the user entirely. If you are in agreement, let's modify this file to remove the exports.

src/mlte/measurement/external_measurement.py
```diff
@@ -0,0 +1,31 @@
...
+        """
+        Initialize a new ExternalMeasurement measurement.
+
+        :param identifier: A unique identifier for the instance
+        :type identifier: str
+        :param result_type: The type of the Result this measurement will return.
+        :type result_type: Type
+        """
+        super().__init__(self, identifier)
+        if not issubclass(Result, result_type):
```
turingcompl33t, line 22: Thank you for adding this explicit runtime type-check!

turingcompl33t (approved):
This is great @sebastian-echeverria. I made one comment about removing `MeasurementMetadata` and `Identifier` from `measurement`-level exports, but I realize that will require a refactor of a lot of the unit tests. We can add that as an issue and tackle it as a separate PR.

src/mlte/measurement/external_measurement.py
sebastian-echeverria: No problem!

Issue #114, opened 2023-03-01
Examine Ability to Reproduce External Measurements


turingcompl33t:
From conversation with @sebastian-echeverria, we determined that we currently have a large gap: with results produced by "internal" (i.e. builtin) measurements, it is trivial to reproduce the result. However, as soon as we introduce the ability to integrate external measurements, it becomes impossible to provide any degree of reproducibility. We simply do not include enough metadata about the external measurement itself to allow users to do this.

The ability to reproduce results should become a primary concern. This is a question that requires further thought and discussion.

PR #116, opened 2023-03-08
Bugfix/artifact path spaces

sebastian-echeverria:
* Fixes issue #114 where spaces in path were being replaced by hyphens, thus breaking access to a path in the artifact store. Now only the name of the file for an artifact has its spaces replaced by hyphen, which was the original intention.
* Added indentation to serialized JSON files in store, to make them more human readable.

### DIFF
--- a/src/mlte/store/api/local/api.py
+++ b/src/mlte/store/api/local/api.py
@@ -100,10 +100,9 @@ def _result_path(model_version_path: Path, result_identifier: str) -> Path:
     :return: The formatted result path
     :rtype: Path
     """
-    path = (
+    return (
         model_version_path / result_identifier.replace(" ", "-")
     ).with_suffix(".json")
-    return Path(str(path))
 
 
 def _spec_is_saved(model_version_path: Path) -> bool:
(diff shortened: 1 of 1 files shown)


src/mlte/store/api/local/api.py
```diff
@@ -100,8 +100,10 @@ def _result_path(model_version_path: Path, result_identifier: str):
     :return: The formatted result path
     :rtype: Path
     """
-    path = (model_version_path / result_identifier).with_suffix(".json")
-    return Path(str(path).replace(" ", "-"))
+    path = (
+        model_version_path / result_identifier.replace(" ", "-")
+    ).with_suffix(".json")
+    return Path(str(path))
```
turingcompl33t, line 106: The type of the above expression should also be `Path`, so there should be no need to cast to `str` and then immediately re-invoke the `Path` constructor. This can be:

```python
return (model_version_path / result_identifier.replace(" ", "-")).with_suffix(".json")
```

turingcompl33t (changes requested):
Small change requested for clarity.

src/mlte/store/api/local/api.py
sebastian-echeverria: Totally agree with that. Note that I didn't add that code, it was already that way. I was wondering why you were repacking that as a Path, and assumed you had a reason, so I prefered not to modify it. But since you don't seem to have a reason for that, I'll change it as you now suggest.

sebastian-echeverria: And I answered my own question: it was that way because the replace didn't work on the Path, but on a string, hence needing the conversion, so it made sense. I've uploaded the updated version.

turingcompl33t (approved):
LGTM

PR #119, opened 2023-03-15
Embed Validator Name in `ValidationResult` Instance

### LINKED ISSUE #118: validator_name field in `ValidationResult` is not Set
The field that records the validator name that produced the instance is not recorded in `ValidationResult`. This is an oversight.

turingcompl33t:
Resolves #118.


krmaffey (approved):
LGTM!

Issue #76, opened 2022-12-13
Support for OSX


turingcompl33t:
Currently, some measurements break on OSX. We need to make the supported platform explicit, and add checks for OSX.

sebastian-echeverria:
This should be closed.

turingcompl33t:
Agreed!

PR #120, opened 2023-03-15
Feature/minor improvements

sebastian-echeverria:
A set of several minor improvements. Includes:
 - Added explicitly separate functions to ProcessMeasurement to start external scripts and general processes (before it was supported underneath, but only the script functionality was enabled under, confusingly, a start_process name that sounded more generic; now both are available under more consistent names).
 - Partially addressed #76 by adding general function to check if running on macOS, and to specifically use it for LocalProcessMemoryConsumption, which does not work on macOS.
 - Minor update to demo: added missing package dependency.
 - Minor update to gitignore to ignore automatically generated coverage files with slightly different filename format.


turingcompl33t (approved):
LGTM!

PR #129, opened 2023-03-22
Refactoring and external results

### LINKED ISSUE #117: Remove Internal Types from Exports
`MeasurementMetadata` and `Identifier` are exported by the `mlte.measurement` module. These are now an implementation detail that should be hidden.

### LINKED ISSUE #127: Reduce Duplication in FS Artifact Store Backend
Local functions that read / write artifacts in the filesystem-based artifact store backend repeat the functionality required to load and store JSON documents. We can collapse these.

### LINKED ISSUE #128: Add Method to `GlobalState` to Check Existence of Required Member Data
We repeat this  logic many times across the project.

sebastian-echeverria:
* Fixes #128 by centralizing check() call as a method of GlobalState
* Fixes #127, refactored to have 1 common read/write artifact internal local api function.
* Fixes #117, making Identifier and MeasurementMetadata internal, adapting imports in tests, internal modules, and extended results to now load them from internal modules. 
* As part of #117, added ExternalMeasurement collect_data() method to create a result of a certain type when there is no function to call/evaluate. In practice, it currently the same as ExternalMeasurement's evaluate(), but it could differ in the future if evaluate() starts really executing the function it is measuring. The main goal of this method is to give the library user a clearer way to indicate it is just wrapping data in the result type provided to the ExternalMeasurement.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,6 @@ demo/models
 demo/store
 demo/media
 demo/reports
+
+# VS Code local configs.
+.vscode/
--- a/demo/confusion_matrix.py
+++ b/demo/confusion_matrix.py
@@ -8,7 +8,7 @@ from typing import Dict, Any
 import numpy as np
 
 from mlte.measurement.result import Result
-from mlte.measurement import MeasurementMetadata
+from mlte.measurement.measurement_metadata import MeasurementMetadata
 from mlte.measurement.validation import (
     Validator,
     ValidationResult,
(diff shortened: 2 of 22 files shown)


sebastian-echeverria:
Addressing several of the things we discussed today. The collect_data() method could have another name. It could also be focused on what it is actually doing, and be called create_result(), for example, or wrap_data(). Let me know your thoughts on this.

turingcompl33t (changes requested):
Two minor comments on naming. It is great how much duplicated code you were able to remove with this PR!

turingcompl33t (approved):
Thanks for the updates, approving and merging.

Issue #145, opened 2023-05-19
Add support for composing Requirements with different logical operators in a Spec


sebastian-echeverria:
Instances of Requirement should be composed via simple logical operations (Or, And, Not?) that in turn produce instances of Requirement.

sebastian-echeverria:
Not sure if this is really needed. Seems that these will turn into Test Cases, and each case will have everything in it. No need to do this sort of composition anymore.

Issue #141, opened 2023-05-11
Add method to load all Values from a store


sebastian-echeverria:
Add a method, probably to the Value class, that allows you to load all Values from the current store into a list of Values. This may be complimentary to the work in #121 , as that could make this easier to do. Alternatively, this could be done without that change in a less elegant way.

turingcompl33t:
Resolved by `Store` query functionality.

sebastian-echeverria:
Even though this is possible by the Store query/filter/search_artifacts functionality, a substantial amount of code is needed to pull it off. It makes more sense to add a frontend, something like Value.load_all(), that underneath does the work to make it easier to load all Values from a Store.

Currently, partial code to do this would look something like this:

```
 with ManagedSession(session().store.session()) as handle:
    query = Query(filter=ArtifactTypeFilter(type=FilterType.TYPE, artifact_type=ArtifactType.VALUE))
    artifacts = handle.search_artifacts("ns", "IrisClassifier", "0.0.1")
    values = [Value.from_model(value_model) for value_model in artifacts]
```

However, this is not complete, as Value.from_model is specific for each Value subtype, so it wouldn't work in this simplified form. Furthermore, this should also work for external Value types extended by the user, so more work needs to be added to support that in a transparent way.

Issue #135, opened 2023-04-23
New `mlte` Workflow


turingcompl33t:
We are introducing a new workflow for `mlte` that touches both requirements definition and validation. @sebastian-echeverria is the primary author for this addition; I am merely adding this comment as a means of recording the ongoing work.

Summary of Terminology Updates:
- `Spec` unchanged
- `Property` unchanged
- `Measurement` unchanged
- `Value` replaces `Result`
- `Condition` replaces `Validator`
- `Result` replaces `ValidationResult`
​
Summary of Implementation Updates:
- A declarative specification (`Spec`) is now the sole means of defining requirements as well as specifying the means by which these requirements will be evaluated
- At the top level, a `Spec` is now a mapping from instances of `Property` to instances of `ReqExpr` (name TBD)
- A `ReqExpr` is the thunk produced by invoking a `Condition` method with its associated threshold (`Conditions` are now higher-order functions that produce functions themselves)
- Instances of `ReqExpr` can be composed via simple logical operations (`Or`, `And`, `Not`?) that in turn produce instances of `ReqExpr`
- Instances of `Measurement` implement a ".value" property method that allows users to access the class type object for the Value produced by the `Measurement`; this supports an ergonomic experience in which developers get autocompletion hints when drafting a Spec
- External `Measurements` are integrated via instances of `ExternalReqExpr` (name TBD) that encode both the "label" attached to the `Value` of interest (the Value produced by the external measurement) as well as the `Condition` that will be applied to this Value

```python
  spec = Spec({
      PredictionAccuracy("prediction accuracy is important because..."): ClassificationAccuracy("label").value.greater_than(0.85),
      InferenceLatency("inference latency should be low"): And(
          LatencyWithCPU("label").value.less_than(100),
          LatencyWithGPU("label").value.less_than(10)
      ),
      PredictionAccuracy("..."): ExternalReqExpr("label", Real.greater_than(0.85))
  })
```

turingcompl33t:
Resolved by #146.

Issue #139, opened 2023-05-03
Rename `MeasurementMetadata`


turingcompl33t:
The `MeasurementMetadata` object tracks information that is relevant to a particular element of evidence throughout its lifecycle. It is not specific to a `Measurement`, or it applies to more than just `Measurement`s. Accordingly, we should update the name to better indicate what it does.

sebastian-echeverria:
This is addressed in #135

turingcompl33t:
Resolved by #146.

PR #146, opened 2023-05-19
Feature/declarative spec part 2

sebastian-echeverria:
Addresses almost all items in issue #135 , the new declarative spec workflow. More specifically:

- Condition was renamed to Requirement.
- Validator was renamed to Condition.
- BoundSpec was renamed to ValidatedSpec. (We can discuss this new name as needed)
- Class method .value() added to Measurements to get the return value they each produce.
- Conditions are generated as thunks that are stored in a Requirement to be validated when a Value is passed to them.
- Modified spec schema to store Condition callbacks as dilled functions with base64 encoding. Arguments were also stored as part of a Condition's json for documentation purposes.
- Modified spec schema to indicate a result including the EvidenceMetadata, to keep track of what measurement type it was generated from.

Things not done:
- Combination of conditions other than an implicit "and" were not added yet. This was moved to new issue #145 .

The Spec creation example from #135 was modified a bit from the issue, and now it would be seen as this:

```
spec = Spec({
      PredictionAccuracy("prediction accuracy is important because..."): [Requirement("label", ClassificationAccuracy().value().greater_than(0.85))],
      InferenceLatency("inference latency should be low"): [
          Requirement("label", LatencyWithCPU().value().less_than(100)),
          Requirement("label", LatencyWithGPU().value().less_than(10))
      ],
      PredictionAccuracy("..."): [Requirement("label", Real.greater_than(0.85))]
  })
```
Note the following changes from the proposal:

- Requirements are explicitly created. This is needed to store the label/identifier properly. In the way it was proposed, the label/identifier could only be stored by the Condition, which was generated by the method that generated the Condition, but we are trying to make Conditions independent of Values/Measurements, so this method allowed breaking that dependency.
- Also, the above makes it almost transparent to create a Requirement from an internal measurement or an external one. Both are created as Requirement, the only difference being that for the first one the value type can be obtained from the specific Measurement class, and for external ones the value type is explicitly written down.
- As mentioned previously, since there is no support for logical combinations of Requirements yet, the way to write them is currently as an array, even if it is an array of only one Requirement (the arrays are an implicit AND).
- Instead of being a property, "value" is a method of Measurement. This is because it doesn't make sense to create an instance of a measurement to get its value type only, and properties are not well supported as class methods in Python, so a simple class method worked better instead.

Some general comments:

- To clarify, currently a Requirement is simply composed of a label/identifier and a Condition. A Condition is an object equivalent to the previous Validators, storing the name of the validation and the arguments for reporting purposes, as well as a callback that has the actual validation function thunk. 
- Please review the schemas for Spec/ValidatedSpec, as they changed a bit. One difference is that the Condition in stored there for each Requirement, including the dilled thunk of the function to validate, which makes it a bit ugly but not that terrible. The JSON for a Condition also includes its name and arguments, mostly for purposes of showing that later in a report, as they can't be extracted easily from the thunk. We can discuss this as needed.
- Another schema change is that the optional "result" item that is only present for a Requirement in the ValidatedSpec (and is the only difference between a ValidatedSpec and a Spec), contains a metadata subsection that includes information on the measurement type it was generated from (before this was at a higher level). This means that, when a Spec is generated, the measurement type to use is not stored in the Spec, only the Condition, and only after gathering the data and validating the conditions we get information about the Measurements used in the ValidatedSpec. This makes sense, but if we want to store the measurement type with the Spec, we can change this. We can discuss this as needed.
- SpecValidator was maintained as an external class used to validate the Spec, to leave all validation functionality out of Spec. Check the report Jupyter Notebook to see how it works. Issue #141 was created to potentially add a method to load all values from a store and make that validation flow a bit easier.


sebastian-echeverria:
Note: this also addresses #139 , by renaming MeasurementMetadata to EvidenceMetadata.

sebastian-echeverria:
This PR also seems to be addressing #95 by making ValidatedSpec contain a Spec and a list of Results, instead of a JSON document. It also adds to/from JSON functionality to the ValidatedSpec (and to Result by extension).

sebastian-echeverria:
This PR also addresses #90 , since now Conditions contain thunks that are executed later, when doing the actual validation.

sebastian-echeverria:
This PR also addresses #67, with the new way Conditions work.

turingcompl33t:
Leaving this here as a reminder for either of us: we must add `dill` to the `pyproject.toml` (or `setup.cfg`) to ensure it is included as a runtime dependency (installed with the package).

src/mlte/evidence/evidence_metadata.py
```diff
@@ -0,0 +1,60 @@
+"""
+EvidenceMetadata class definition.
```
turingcompl33t, line 2: I like this name much better than `MeasurementMetadata`. Thank you for coming up with something good!

src/mlte/measurement/cpu/local_process_cpu_utilization.py
```diff
@@ -202,6 +205,11 @@ def __call__(self, pid: int, poll_interval: int = 1) -> CPUStatistics:
             max=max(stats),
         )
 
+    @classmethod
+    def value(self) -> Type[CPUStatistics]:
```
turingcompl33t, line 209: I am wondering if there is a way to automate this through some additional runtime introspection? Such that each measurement we implement doesn't need to provide a similar method that provides the class type object for the value that it produces. I might experiment with this a bit.

src/mlte/property/property.py
```diff
@@ -46,12 +46,12 @@ def __init__(self, name: str, description: str, rationale: str):
...
         """The rationale for using the property."""
 
-    def _to_json(self) -> Dict[str, str]:
+    def _to_json(self) -> dict[str, str]:
         """
         Save a Property instance to a JSON document.
 
         :return: The document
-        :rtype: Dict[str, str]
+        :rtype: eict[str, str]
```
turingcompl33t, line 54: Small typo.

src/mlte/spec/requirement.py
```diff
@@ -0,0 +1,82 @@
...
+        )
+
+    def validate(self, value: Value) -> Result:
+        """
+        Validates if the given value matches the requirement.
+
+        :return: The result of validating this requirement.
+        :rtype: Result
+        """
+        result: Result = self.condition(value)
```
turingcompl33t, line 60: This is pretty awesome how simple validation becomes now that you have all the machinery in place.

src/mlte/spec/spec_validator.py
```diff
@@ -32,79 +31,52 @@ def __init__(self, spec: Spec):
...
-        # Check that all conditions have values to be validated.
-        conditions = self.spec.conditions[property.name]
-        for condition in conditions:
-            if condition.label not in values:
-                raise RuntimeError(
-                    f"Condition '{condition.label}' does not have a value that can be validated."
+        # Check that all requirements have values to be validated.
+        for _, requirement_list in self.spec.requirements.items():
+            for requirement in requirement_list:
+                if str(requirement.identifier) not in self.values:
```
turingcompl33t, line 70: This has always been an open question for me: do we allow users to validate with insufficient evidence? Do we generate a report that shows that certain properties are not covered by the values collected? I am not sure which is desirable from a user perspective.

I have also considered enabling the above functionality, but then also included a `strict` flag on the validation process that fails immediately if there is insufficient evidence.

src/mlte/validation/condition.py
```diff
@@ -0,0 +1,119 @@
...
+        """        try:
+            validator = getattr(value, self.condition)
+        except AttributeError:
+            raise RuntimeError(
+                f"Invalid validation method provided: '{self.condition}()' method not found for value of type {value.typename}"
+            )
+        condition: Condition = validator(self.threshold)"""
+        return {
+            "name": self.name,
+            "arguments": self.arguments,
```
turingcompl33t, line 69: Perhaps I missed this elsewhere in the review, but what are these `arguments`? Are we sure that they can be serialized?

turingcompl33t (approved):
I think this looks good to merge, @sebastian-echeverria. No problem with deferring the ability to compose measurements to a separate PR down the road. I left some comments, but they are just questions rather than issues that need to be addressed.

src/mlte/property/property.py
sebastian-echeverria: Fixed!

src/mlte/spec/spec_validator.py
sebastian-echeverria: That is a very good question. I am being more strict now by default, but I am not sure what the best option is. Maybe the strict flag would be a good compromise. Created #149 for this.

src/mlte/measurement/cpu/local_process_cpu_utilization.py
sebastian-echeverria: It doesn't seem to be very straightfoward, but I guess it could be possible through instrospection of the __call__ method.

src/mlte/validation/condition.py
sebastian-echeverria: That is a very good question. I mentioned it briefly above, but these are the one or many arguments passed over to the condition thunk when validating. Typically it is the threshold. These arguments are stored in the callback, and as thus are implicitly stored in the thunk that is dilled into the JSON file. So this field is not really required. My main reasoning to add it was to leave documentation of the arguments passed into the condition, since they end up being hidden in the callback, and there is no easy way to recover them and see what was passed to it after loading a Spec. So if a condition is .less_than(3), the name of the condition, "less_than", is stored in the spec, but the "3" is only stored dilled in the callback, so there is no way to see it. For a user to be able to see "less_than" what, we need to store it in a separate field. This field is actually not used in any way other than maybe to show this potentially in a UI, or report, or when examining the Spec. So I believe it is important to have it.

sebastian-echeverria: That being said, I am not sure they can be serialized. In practice, the arguments we have for the current validators are always primitives to they are serializable, but in theory new conditions could use non-serializable arguments. I opened ticket #150 for this.

src/mlte/measurement/cpu/local_process_cpu_utilization.py
turingcompl33t: Not a blocking comment.

src/mlte/property/property.py
turingcompl33t: Resolved.

src/mlte/spec/spec_validator.py
turingcompl33t: Great, thanks!

PR #167, opened 2023-07-12
Simplifying addition of new properties

sebastian-echeverria:
Two minor changes:
 - __repr__ was moved to the base Property class, since it works there anyway, so creating a new Property does not require one to duplicate that function.
 - When loading a Property from a JSON file, removed hardcoded list of subpackages containing properties, and changed to use reflection to find all subpackages containing properties, to simplify the addition of properties in new subpackages.


turingcompl33t (approved):
LGTM.

PR #169, opened 2023-07-20
Feature external measurement traceability

sebastian-echeverria:
Partially addressed #114 to add traceability to ExternalMeasurements. For now, it is able to store information about the function used for the evaluation, though this is not enough to reproduce, but it helps with traceability.

Specific changes:

- ExternalMeasurement now gets an additional optional parameter in its constructor, which is the function to be called when evaluating. It is optional if evaluate() will not be used, such as when ExternalMeasurement is used only for ingest() to create a Value wrapper.
- eval() in ExternalMeasurement now executes the function received in the constructor, passing the arguments it receives. This is closer to how built-in measurements work, in that the Measurement itself executes the measurement, instead of the previous behavior, where it was just receiving the result of that measurement.
- EvidenceMetadata was modified to add an additional optional argument, additional_info. This unstructured string can be use to add additional metadata. It was left generic to avoid adding a soft dependency to ExternalMeasurements, and allow EvidenceMetadata to use that attribute more generically. 
- When created, ExternalMeasurement, if it received a function, will add a string with the function module and function name, along with the prefix "function: ", to the additional_info of the EvidenceMetadata it creates and attaches to future Values resulting from measurements.

JSON for EvidenceMetadata with additional info for a function now looks like this:
```
      "metadata": {
          "measurement": {
              "identifier": "accuracy",
              "measurement_type": "ExternalMeasurement",
              "additional_info": "function: sklearn.metrics._classification.accuracy_score"
          },
          "value": {
              "typename": "Real"
          }
      },
```


src/mlte/evidence/evidence_metadata.py
```diff
@@ -40,14 +48,24 @@ def from_json(json: dict[str, Any]) -> EvidenceMetadata:
             raise RuntimeError(
                 "Cannot deserialize EvidenceMetadata, missing key 'measurement_type'."
             )
-        return EvidenceMetadata(
+
+        additional_info: Optional[str] = None
```
turingcompl33t, line 52: Nit: ternary logic:

```python
additional_info: Optional[str] = doc["additional_info"] if "additional_info" in json else None
```

src/mlte/measurement/external_measurement.py
```diff
@@ -4,32 +4,59 @@
...
 
 
 class ExternalMeasurement(Measurement):
-    def __init__(self, identifier: str, value_type: type):
+    @typing.no_type_check
+    def __init__(
+        self,
+        identifier: str,
+        value_type: type,
+        function: Optional[Callable] = None,
```
turingcompl33t, line 20: Question: we can't annotate the type for this callable any further, correct? The signature of the callable is totally opaque to us?

src/mlte/measurement/external_measurement.py
```diff
@@ -4,32 +4,59 @@
...
+
         if not issubclass(Value, value_type):
             raise Exception(
-                f"Value type provided is not a subtype of Value: {self.value_type}"
+                f"Value type provided is not a subtype of Value: {value_type}"
             )
         self.value_type: type = value_type
 
+        if function is not None:
+            if not callable(function):
```
turingcompl33t, line 41: Info: this is a cool function, I didn't know this was a builtin.

src/mlte/evidence/evidence_metadata.py
sebastian-echeverria: Ok, good point. I'll add it in another branch.

src/mlte/measurement/external_measurement.py
sebastian-echeverria: Yeah, that is the point. Just a function we call that can be anything they want. I don't see how to further restrict it since what we want is to give them freedom there.

src/mlte/measurement/external_measurement.py
sebastian-echeverria: Just learned it myself :)

Issue #174, opened 2023-08-03
Add Paper Citation in Top-Level README


turingcompl33t:
We should cite the MLTE [paper](https://arxiv.org/abs/2303.01998) in the top-level README for the repository. See the [cleverhans](https://github.com/cleverhans-lab/cleverhans#citing-this-work) repository README for an example of the format we might use (we should include a citation in the README).

PR #175, opened 2023-08-03
Add paper to README.md

krmaffey:
Resolves issue #174

@turingcompl33t, a note - it appears that the conference version of our paper is not posted publicly anywhere online, so I cited the arXiv version. We may want to change that at some point.

Please check the language I used to describe MLTE, because I updated it from 'toolkit'. 

Also, hilarious that you used cleverHans as an example because before I put that comment into the doc, I checked their repository to see if I had remembered correctly that they cited their paper. It's a great example.


turingcompl33t (approved):
Thanks @krmaffey, LGTM.

Issue #182, opened 2023-08-10
Implement Filesystem Backend for Artifact Store


turingcompl33t:
Currently, we have only `InMemoryStore` and `RemoteHttpStore` as underlying representations for the MLTE artifact store. We need a true persistence mechanism for any non-toy use. A filesystem-based store implementation is the simplest, and one that we supported in the past. We should port this work to a new `LocalFilesystemStore` implementation.

turingcompl33t:
Resolved by #190.

PR #190, opened 2023-08-14
Feature fs store

sebastian-echeverria:
This PR addresses #182 to add a local File System store. It works similarly to how the original file system store worked like. Some overall comments on the implementation:

- It stores namespaces, models and versions as nested folders inside the provided store folder.
- It stores artifacts as JSON files inside the version folders.

There are couple of other minor fixes around, mostly comments and formatting.

Some notes on things that were not explicitly included here:
- The file store can not currently used by any artifact other than the NegotiationCard, as other artifacts have not been updated yet to use the new store system.
- There are no explicit checks to see if a namespace, model, version or artifact id/name contains characters that are invalid to be file/folder names. (Not sure best way to handle this... it seems that the previous FS store was not handling much of this either).

The implementation was modeled following the RemoteHTTP and Memory stores, but I am not 100% sure I understood all the underlying assumptions in the current design (exact difference between Store and Storage, and SessionStore, etc), so there may be inconsistencies with the expected design structure.


src/mlte/store/underlying/fs.py
```diff
@@ -0,0 +1,458 @@
...
+    """Removes a folder and its contents."""
+    for p in root.iterdir():
+        if p.is_dir():
+            _rmtree(p)
+        else:
+            p.unlink()
+    root.rmdir()
+
+
+class JsonFileStorage:
```
turingcompl33t, line 51: This is a nice wrapper for some of the internal functionality.

test/store/underlying/test_local_fs.py
```diff
@@ -0,0 +1,193 @@
+"""
```
turingcompl33t, line 1: I appreciate your bringing in all of these tests. However, in my latest PR (that has since merged) I refactored tests for the underlying store representation to be parametric over the backend. That means that it should be possible to remove this file, add a fixture in `test/store/fixture.py`, and update the configuration in that file, and all of the existing artifact store tests should now execute for the filesystem implementation as well.

turingcompl33t (commented):
Few comments, one of which may further simplify this diff, and another which may require slight modifications to the implementation. Overall, this is great. Easy to follow implementation.

src/mlte/store/underlying/fs.py
```diff
@@ -0,0 +1,458 @@
...
+    :rtype: Path
+    """
+    assert uristr.startswith("fs://") or uristr.startswith(
+        "local://"
+    ), "Broken precondition."
+    prefix = "fs://" if uristr.startswith("fs://") else "local://"
+    return Path(uristr[len(prefix) :])
+
+
+def _rmtree(root: Path):
```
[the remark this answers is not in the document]
sebastian-echeverria: I believe we can. Since we were using the pathlib library for everything (which, in the past, I have not used that frequently), I was trying to look for an implementaiton that would only use Pathlib as well. But if you don't have a preference, I can change to shutil.rmtree()

test/store/underlying/test_local_fs.py
sebastian-echeverria: Yes, to be honest, I noticed that they were the same tests for the other stores, and I remembered you had talked about making them parametric, but I guess I accidentally didn't merge that last version of master and I thought that was not done yet. I'll do that and update this.

src/mlte/store/underlying/fs.py
```diff
@@ -0,0 +1,458 @@
...
+        :param model_id: The identifier for the model
+        :param version_id: The identifier for the version
+        :raises ErrorNotFound: If the required structural elements are not present
+        :return: The associated artifacts
+        """
+        self._ensure_namespace_exists(namespace_id)
+        self._ensure_model_exists(namespace_id, model_id)
+        self._ensure_version_exists(namespace_id, model_id, version_id)
+
+        return [
```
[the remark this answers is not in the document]
sebastian-echeverria: That is not something that I really looked into. But I'll take a quick look and see if I can add it.

sebastian-echeverria:
Ok, addressed the 3 comments above. Using shutil now, using sorting for the files/artifacts, and merged and updated store tests to use the fixtures and parametric tests.

test/store/fixture.py
```diff
@@ -75,6 +76,12 @@ def memory_store() -> InMemoryStore:
     return InMemoryStore(StoreURI.from_string("memory://"))
 
 
+@pytest.fixture(scope="function")
```
turingcompl33t, line 79: Perfect.

src/mlte/store/underlying/fs.py
```diff
@@ -0,0 +1,459 @@
...
+    def list_folders(self, path: Path) -> list[Path]:
+        return [x for x in sorted(path.iterdir()) if x.is_dir()]
+
+    def delete_folder(self, path: Path) -> None:
+        shutil.rmtree(path)
+
+    def list_json_files(self, path: Path) -> list[Path]:
+        return [
+            x
+            for x in sorted(path.iterdir())
```
turingcompl33t, line 65: Perfect.

turingcompl33t (approved):
LGTM.

PR #191, opened 2023-08-15
Refactor Values to Artifact Protocol

### LINKED ISSUE #156: Refactor Value to Artifact Protocol
Refactor the implementation of `Value` to comply with the MLTE artifact protocol.

	- `to_model()`
	- `from_model()`
	- `save_with()`
	- `load_with()`

turingcompl33t:
Resolves #156.

This PR refactors the MLTE `Value` system to the artifact protocol. Specifically, it makes the following changes:

- Refactors MLTE value types to be compatible with the artifact protocol
- Refactors the artifact protocol itself slightly to utilize discriminated unions; this was not an issue in the original PR because there was only a single artifact type
- Refactors the implementation of `EvidenceMetadata` to be serializable, like a MLTE artifact


turingcompl33t:
Remaining items:

- Refactor the `EvidenceMetadata` and `Identifier` classes to exist in the same module
- Implement support for extending `Values` with the `mlte.value.base` implementation

demo/confusion_matrix.py
sebastian-echeverria, line 1: Even if we are not updating the whole demo yet, I think it would be a good idea to update the ConfusionMatrix to use the new ArtifactModel structure, to showcase how this would work for Values that are extended by users.

src/mlte/artifact/artifact.py
sebastian-echeverria, line 1: The fact that save_with and load_with can now be done in the base Artifact is very good news, as it simplifies all other Artifacts. I guess this comes from creating the read and write artifact functions instead of the specific read/write for each artifact we had before, that you did at the end of the previous PR.

test/fixture/artifact.py
```diff
@@ -50,14 +55,19 @@ def all_others(type: ArtifactType) -> list[ArtifactType]:
...
     """
     Make the body of the artifact for a given type.
     :param type: The artifact type
+    :param id: The identifier for the artifact
     :return: The artifact body model
     """
     if type == ArtifactType.NEGOTIATION_CARD:
         return _make_negotiation_card()
+    if type == ArtifactType.VALUE:
+        return _make_value(id)
```
sebastian-echeverria, line 70: I guess this is another step now for adding new artifacts, adding it to this list and creating the fixture? Or rather, this replaces having to create the unit tests in part, to use the parameterized ones for common tests.

sebastian-echeverria (changes requested):
The PR looks good, and I think the improvements to the Artifact as well as the internal ones make sense. I would wait for you to merge this before doing much work on the Spec artifact, though I'll start in parallel, since there are several things here that will be needed or are improvements to be used when porting Spec as an Artifact. I only have one comment below about the ConfusionMatrix as a useful example to include in this PR.

demo/confusion_matrix.py
turingcompl33t: I added the `mlte.value.base` implementation this morning that supports extending the MLTE value system with user-defined types. There is now a unit test (`test/value/test_extension.py`) that shows how a `ConfusionMatrix` type can be implemented as an extension.

src/mlte/artifact/artifact.py
turingcompl33t: Indeed!

test/fixture/artifact.py
turingcompl33t: That is true, I will add this to the list of items on the remaining artifact protocol port issues.

demo/confusion_matrix.py
turingcompl33t: I will update the demo accordingly as well.

turingcompl33t:
Remaining items:
- Refactor the EvidenceMetadata and Identifier classes to exist in the same module
- Add "update `test/fixture/artifact.py` to TODO list for artifact protocol ports
- Update the `ConfusionMatrix` type in demo

src/mlte/value/base.py
```diff
@@ -0,0 +1,111 @@
...
+from mlte.evidence.metadata import EvidenceMetadata
+from mlte.value.model import OpaqueValueModel, ValueModel, ValueType
+
+
+def _has_callable(type, name) -> bool:
+    """Determine if `type` has a callable attribute with the given name."""
+    return hasattr(type, name) and callable(getattr(type, name))
+
+
+class Value(Artifact, metaclass=abc.ABCMeta):
```
sebastian-echeverria, line 33: This looks mostly great, but I do have one question. I understand what you did, adding the abstract seriaize/deserialize methods for extensions to implement, and implementing to/from moodel by using Opaque and calling the abstract serialize/deserialize that an extended value should implement. All of that looks fine and makes sense. What I do not understand is why you created a new Value class, with the same name as the Value artifact, and copying/pasting all the code from the Value artifact class, instead of deriving from it. Again, I try to avoid having duplicated code when possible. Was it just so it could be called Value too? Why couldn't this be called something like BaseValue, and have it extend Value, and simply add the 2 abstract methods and the 2 implemented methods? Is it to avoid some of the polymorphism/model issues?

turingcompl33t: The primary reason for this is the "contract" that the client of the type agrees to. We (developers of MLTE internals) override the `mlte.value.artifact` type; to do this, we have to provide `to_model()` and `from_model()` methods to satisfy the artifact protocol. On the other hand, external users of MLTE override the `mlte.value.base` type; to do this, they have to provide the `serialize()` and `deserialize()` methods. Moreover, this latter type already provides the `to_model()` and `from_model()` methods required by the artifact protocol. Ultimately, these two types exist at slightly different levels of abstraction.

I suppose that we could have this derive from the `Value` artifact, like you suggested, and call it `BaseValue`. The reason I didn't take this approach was because I don't want the user to ever have to know about the distinction between `artifact.value` and `base.value`. They are only ever exposed to `base.value` and don't have to worry about the extra work that we have to do to make the abstraction work.

sebastian-echeverria: Ok, that is fine, I understand and agree with having the two Values. I also understand the reasoning for keeping its name Value, and that is fine too. But we could still derive Value from Value, and keep them with the same name, right? We would just need, inside base.value, to use full package names or alias to derive it from artifact.value. This would still be transparent to the user, but simplify maintainance effort by not duplicating the common code between base.value and artifact.value. This would still be hidden to the user.

sebastian-echeverria (commented):
Please see my comment about still trying to derive Value from Value without changing either's name. Other than that I am ok with this PR.

src/mlte/value/base.py
turingcompl33t: Agreed. I just tried this approach, and it does indeed work to have `base.Value` derive from `artifact.Value`. However, the only code duplication this removes are a couple of lines in the `__init__` method in `base.Value`. Other than that, there is little overlap between the two. The major implementation in `base.Value` are the `to_model()` and `from_model()` methods. These are not implemented by the `artifact.Value` class. Are there other bits of functionality that you are concerned about? Otherwise, I don't think its worth extending the inheritance hierarchy one level deeper to remove these three lines in the constructor (because it makes it slightly harder to reason about the types, in my opinion).

sebastian-echeverria (approved):
LGTM

src/mlte/value/base.py
turingcompl33t: Resolved in Slack.

PR #198, opened 2023-08-17
Support Implicit Creation of Organizational Elements on Artifact Write

### LINKED ISSUE #195: Artifact Store `write_artifact()` Should Support Creation of "Parents"
In previous versions of the MLTE artifact store we supported implicit creation of organization elements for an artifact. For instance, if we saved an artifact with a particular namespace, model, and version, these elements would be implicitly created if they did not already exist. In the refactor of the API, I removed this functionality.

After conferring with @sebastian-echeverria, we have decided to add this functionality back in, in an opt-in manner. This will entail the following:

- Update the API for `StoreSession.write_artifact` to accept a `parents` argument. This argument will default to `False`. When `True`, organizational elements for the artifact will be implicitly created, if they do not already exist.
- Update the web API for artifact read and write requests. This will allow us to provide "parameters" to the request, like `parents`, that are distinct from the artifact itself.
- Propagate this functionality up to the `Artifact` interface, such that `parents` can be specified on `Artifact.save()`

turingcompl33t:
Resolves #195.

This PR adds the ability to implicitly create organizational elements on write to the artifact store. Three distinct interfaces have been updated:

- The `Store` interface now supports an optional parameter, `parents`, to `write_artifact`, that specifies that organizational elements should be created if they do not exist
- The web API endpoint for artifact write operations has been modified to support specifying `parents` in the request
- The artifact protocol base type, `Artifact`, now supports `parents` in `Artifact.save()` and `Artifact.save_with()`

Ultimately, this means that we can now write code like the following:

```python
card = NegotiationCard("my-card")
card.save(parents=True)
```

And the write will succeed, even when the namespace, model, and version were not previously created for the artifact.

### DIFF
--- a/src/mlte/negotiation/negotiation_card.py
+++ b/src/mlte/negotiation/negotiation_card.py
@@ -12,7 +12,7 @@ import mlte.negotiation.model as model
 from mlte.artifact.artifact import Artifact
 from mlte.artifact.model import ArtifactHeaderModel, ArtifactModel, ArtifactType
 from mlte.context.context import Context
-from mlte.store.store import ManagedSession, Store
+from mlte.store.base import ManagedSession, Store
 
 
 class NegotiationCard(Artifact):
--- a/src/mlte/negotiation/negotiation_card.py
+++ b/src/mlte/negotiation/negotiation_card.py
@@ -58,11 +58,15 @@ class NegotiationCard(Artifact):
             model=model.body.model,
         )
 
-    def save_with(self, context: Context, store: Store) -> None:
+    def save_with(
+        self, context: Context, store: Store, *, parents: bool = False
+    ) -> None:
         """
         Save an artifact with the given context and store configuration.
         :param context: The context in which to save the artifact
         :param store: The store in which to save the artifact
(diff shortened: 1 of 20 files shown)


src/mlte/negotiation/negotiation_card.py
```diff
@@ -58,18 +58,23 @@ def from_model(model: ArtifactModel) -> NegotiationCard:  # type: ignore[overrid
...
+        :param parents: Indicates whether organizational elements for the
+        artifact are created implicitly on write (default: False)
         """
         with ManagedSession(store.session()) as handle:
             handle.write_artifact(
                 context.namespace,
                 context.model,
                 context.version,
                 self.to_model(),
+                parents=parents,
```
sebastian-echeverria, line 77: I guess this will be moved to Artifact with the Value Artifact PR, but I understand why this is here for now.

sebastian-echeverria (approved):
Looks good to me! This will be very helpful.

src/mlte/negotiation/negotiation_card.py
turingcompl33t: Correct!

PR #201, opened 2023-08-17
Refactor Dependency Management

### LINKED ISSUE #199: Refactor Dependency Management
Currently, the dependency management system utilizes different sets of dependencies for different versions of Python. Furthermore, we do not pin versions explicitly, instead we allow them to be computed by `pip-tools`. We need to refactor this approach to make development stable. Different versions of important dependencies now have an impact on the library APIs and even on some of the artifacts that are generated in the MLTE source (e.g. artifact schemas).

turingcompl33t:
Resolves #199.


turingcompl33t:
@sebastian-echeverria I ran into a bunch of issues recently while developing related to dependency management. It stems from the fact that I was not pinning our dependency versions, so they were allowed to float and change with the Python version. In this PR, I migrated us to a Poetry-based build and tried to make dependency management more-sane. Let me know if you have issues with migrating from `setuptools` to Poetry, but I think this will really be a big step towards making the package portable again.

turingcompl33t:
- This PR is based on the `Value` PR; I do plan to merge that one first before merging this one.
- I moved everything up from the `src` directory because this seems to be more common in Python package structure now. More importantly, the Poetry tool appears to operate more effectively with the package source in this directory. With the `src` layout, I can't get the tool to install the MLTE package locally, in an editable mode, which is a nice feature for CLI development.

sebastian-echeverria (dismissed):
Sorry to see that the "from __future__ import annotations" line was not enough to get dict/list to work in 3.8. I am ok with all the changes, or that it had downsides enough for you to undo that. In any case, all changes look good to me.

krmaffey (approved):
Just a fun, casual addition of 1500 lines of code!

PR #207, opened 2023-08-19
Address Bad Updates from Merge

### LINKED ISSUE #206: Address Merge Conflicts from Recent Activity
@sebastian-echeverria pointed out that two recent updates I made were lost recently while resolving merge conflicts:

- `NegotiationCard` should not define `save_with` and `load_with`
- The `parents` parameter is not propagated in `Artifact`

This PR should also introduce a unit test that ensures that tests will fail if the latter bug is present.

turingcompl33t:
Resolves #206.


sebastian-echeverria (approved):
Looks good!

PR #208, opened 2023-08-20
Unify the Implementation of Metaclass Functions

### LINKED ISSUE #202: Generalize Basic Support for Class Derivation Checks
In many places, we use ad-hoc code to ensure that deriving classes implement required methods. We should pull this code out into some generic functionality and resuse it across implementations.

turingcompl33t:
Resolves #202.

In addition, it adds a unit test for `mlte.value.base.Value` to ensure that custom implementations that do not meet the interface cannot be instantiated.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,8 +79,7 @@ check: check-format check-lint check-typecheck
 # Unit Tests
 # -----------------------------------------------------------------------------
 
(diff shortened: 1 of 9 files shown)


sebastian-echeverria (approved):
Looks good!

PR #223, opened 2023-08-22
Implement Versioning Semantics for Artifact Store

### LINKED ISSUE #171: Determine Semantics for Versioning in Artifact Store
Since the inception of the artifact store, I have been thinking that we would want to support implicit multiversioning for artifacts. However, it is not clear to me now whether this is the right approach. The current storage backend we have implemented (memory) does not support versioning. We should determine the semantics we want and implement this accordingly.

### LINKED ISSUE #178: Implement Versioning Semantics in Artifact Store
Once #171 is resolved, we must implement the desired semantics. Currently, single-versioning with error on collision is implemented. If the desired semantics differ from these, we must refactor.

turingcompl33t:
- Resolves #171 
- Resolves #178 

We determined that the current way forward with artifact versioning and write semantics is as follows:

- By default, attempts to overwrite existing artifacts fail with `ErrorAlreadyExists`
- Clients can opt-in to overwrite behavior with the use of optional parameters in MLTE clients

This behavior is important for interactions in the web UI. Here, users will be interested in iterative development of the content in specific artifacts, like the negotiation card and the report. We don't want users to have to create a new version of the artifact every time they enter some new text and hit "save." Therefore, the frontend will implicitly use this overwrite functionality when an existing artifact is edited.

This functionality is exposed by the artifact protocol. It is now possible to write artifacts that overwrite existing instances as follows:

```python
card = NegotiationCard("my-card")
card.save(force=True)
```

### DIFF
--- a/mlte/artifact/artifact.py
+++ b/mlte/artifact/artifact.py
@@ -57,25 +57,34 @@ class Artifact(metaclass=abc.ABCMeta):
             "Artifact.from_model() not implemented for abstract Artifact."
         )
 
-    def save(self, *, parents: bool = False) -> None:
+    def save(self, *, force: bool = False, parents: bool = False) -> None:
         """
         Save an artifact with parameters from the configured global session.
 
         This is equivalent to calling:
             artifact.save_with(session().context, session().store)
 
+        :param force: Indicates that an existing artifact may be overwritten
         :param parents: Indicates whether organizational elements for the
         artifact are created implicitly on write (default: False)
         """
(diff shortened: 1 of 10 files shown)


krmaffey (approved):
Looks good Kyle! My only note is, would it be useful add something to ErrorAlreadyExists to tell the user that they can force an overwrite? Or do you think the parameter descriptions are enough for them to figure it out?

sebastian-echeverria:
Looks ok to me, at least for now.

krmaffey (approved):
Change looks good

PR #234, opened 2023-08-22
Spec Refactoring to Artifact Protocol

### LINKED ISSUE #155: Refactor Spec to Artifact Protocol
Refactor the implementation of `Spec` to comply with the artifact protocol.

	- `to_model()`
	- `from_model()`
- Update `test/fixture/artifact.py` to include the artifact in existing, parametric unit tests

### LINKED ISSUE #160: Experiment with Syntax Refactor for Spec
Determine if it is necessary to introduce the new `Requirement` type to the specification, or if it is possible to inject the required metadata directly into the condition without introduced this additional element of the hierarchy.

### LINKED ISSUE #181: Implement Syntax Refactor for Spec
#160 describes a proposal for a slightly different design for the definition of `Spec` artifacts. Once this design is known, we must implement it. This issue represents that work.

sebastian-echeverria:
- Resolves #155 
- Resolves #160
- Resolves #181 

This PR focuses on creating a model for Spec, and updating it to use the new Artifact design. Some notes:

- There is a strong connection between Spec and ValidatedSpec, but ValidatedSpec will be handled separately in a next PR (issue #157)
- While updating Spec, the Requirement construct was removed, as it was not needed to wrap ids, Conditions and Results, and for now a simpler Dict was enough. This modifies slightly how the user defines a Spec with code, but not much. Some further modifications will be needed when updating ValidatedSpec, to properly associate results with conditions.
- For now, the repr field of Properties was removed, as it was currently redundant with Name and was just adding complexity.

Creating a Spec now looks like this:

```
spec = Spec("spec", {
    TaskEfficacy("Important to understand if the model is useful for this case"): 
                    {"accuracy": Real.greater_or_equal_to(0.98),
                     "confusion matrix": ConfusionMatrix.misclassification_count_less_than(2),
                     "class distribution": Image.ignore("Inspect the image.")},
    StorageCost("Critical since model will be in an embedded decice"): 
                    {"model size": LocalObjectSize.value().less_than(3000)},
    TrainingMemoryCost("Useful to evaluate resources needed"): 
                    {"training memory": LocalProcessMemoryConsumption.value().average_consumption_less_than(0.9)},
    TrainingComputeCost("Useful to evaluate resources needed"): 
                    {"training cpu": LocalProcessCPUUtilization.value().max_utilization_less_than(5.0)}
    })
```


mlte/property/property.py
```diff
@@ -35,91 +36,60 @@ def __init__(self, name: str, description: str, rationale: str):
...
-    def _to_json(self) -> Dict[str, str]:
+    def to_model(self) -> PropertyModel:
         """
-        Save a Property instance to a JSON document.
+        Return a Property as a model.
 
-        :return: The document
-        :rtype: Dict[str, str]
+        :return: The property as its model.
+        :rtype: PropertyModel
```
turingcompl33t, line 51: Just for your knowledge: I switched our documentation generator from `sphinx` to `mkdocs`. The API reference generator for `mkdocs` can infer types for docstrings from Python type annotations, so we no longer need to annotate the parameter types and return types explicitly. This is not a big deal so I won't comment on it any further; eventually we'll just go through and remove the explicit annotations.

mlte/property/property.py
```diff
@@ -35,91 +36,60 @@ def __init__(self, name: str, description: str, rationale: str):
...
-        submodule = importlib.import_module(
-            submodule_info.name, package=properties_package_name
+        if model.name == "":
+            raise RuntimeError(
+                "Property is malformed, it does not have a valid name."
+            )
+        classname = model.name
+
+        # Load the class type from the module
+        properties_package_name = "mlte.property"
```
turingcompl33t, line 77: This is likely similar to the machinery that we might add to `Value.from_model()` to dynamically infer the value type.

mlte/spec/__init__.py
```diff
@@ -1,6 +0,0 @@
-from .requirement import Requirement
```
turingcompl33t, line 1: It looks like you caught on that I am trying to get rid of some of these re-exports from sub-packages. I know it makes things slightly more verbose, but it has solved so many import issues across the package that I think it is worth it. Thanks!

mlte/spec/spec.py
```diff
@@ -35,200 +36,131 @@ def _unique(collection: List[str]) -> bool:
...
+        if not _unique([p.name for p in self.properties.keys()]):
             raise RuntimeError("All properties in Spec must be unique.")
 
         if not _unique(
-            [
-                str(requirement.identifier)
-                for _, req_list in requirements_by_property.items()
-                for requirement in req_list
-            ]
+            list(
```
turingcompl33t, line 63: This is a beautiful expression, but is this true? Should it not be possible to utilize the same "evidence" to validate more than one property? I know this is something of a philosophical question, but I don't see an immediate reason to prevent this.

mlte/validation/condition.py
```diff
@@ -110,6 +102,7 @@ def __str__(self) -> str:
 
     def __eq__(self, other: object) -> bool:
         """Compare Condition instances for equality."""
+        # TODO: is just names enough? Should we compare args and callback?
```
turingcompl33t, line 105: Good question.

turingcompl33t (approved):
This is wonderful. I left only a single question in the PR, but it is not something that prevents merging. LGTM.

mlte/property/property.py
sebastian-echeverria: Correct, this could be a good starting point. The main complication with Values is that extended values can be in any package, so the handling of the full path will be needed, and the code will be a bit trickier.

mlte/spec/__init__.py
sebastian-echeverria: Yes, trying to follow that convention. It is a bit more verbose, but actually I'm used to doing it that way, and getting rid of the circular import issues makes it really worth it.

mlte/spec/spec.py
sebastian-echeverria: That is a good question. This isn't something I just added, btw, it was there from the previous implementation with Requirements, just in a slightly different form, but the question is still valid. Right now, there is an implementation issue with removing this. Not in the Spec itself, but when validating the Spec. In the `SpecValidator`, results are stored by measurement_id, but not separately by property, which means that two conditions using the same `Value` (with the same measurement id) would end up overwriting one another in the list of results. This is more of a limitation of the SpecValidator than a high level requirement, though. I can look into this, since I think refactoring SpecValidator a bit to support this should be simple, but I would prefer to do it in the subsequent PR, #237, which 1) deals more with the validation aspects, and 2) is based on this PR.

mlte/validation/condition.py
sebastian-echeverria: I guess the answer is, we probably should? But maybe it makes sense to expand this once we tackle the adding a registry of the Value that generated the Condition as well, since that would also expand the comparison anyway. I'll leave a note in that new issue.

PR #236, opened 2023-08-22
Report Relative Paths on List Operations for Filesystem Store Backend

### LINKED ISSUE #235: Organizational Elements Should be Listed with Relative Paths
@sei-aderr noticed a bug in the filesystem backend for the artifact store. When organizational elements (namespaces, models, versions) are listed, the root path for the store is included in the name of the element that is returned. For example, if we create the namespace `ns0` and use `store` as the root directory for the filesystem-backed artifact store, then when we perform a list operation on namespaces, the result is:

```
[
  "store/ns0"
]
```

This result should be

```
[
  "ns0"
]
```

turingcompl33t:
Resolves #235.

### DIFF
--- a/mlte/store/underlying/fs.py
+++ b/mlte/store/underlying/fs.py
@@ -121,8 +121,8 @@ class LocalFileSystemStoreSession(StoreSession):
 
     def list_namespaces(self) -> List[str]:
         return [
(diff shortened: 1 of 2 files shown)


sebastian-echeverria (approved):
LGTM

PR #237, opened 2023-08-22
ValidatedSpec Refactoring to Artifact Protocol

### LINKED ISSUE #157: Refactor ValidatedSpec to Artifact Protocol
Refactor the implementation of `ValidatedSpec` to comply with MLTE artifact protocol.

	- `to_model()`
	- `from_model()`
- Update `test/fixture/artifact.py` to include the artifact in existing, parametric unit tests

sebastian-echeverria:
- Resolves #157 

NOTE: This PR is based on #234 , and as such it should be merged after that one. After that one is merged, we may want to refresh this PR so it is easier to see only its separate changes.

This PR focuses on creating updating ValidatedSpec to use the new Artifact design. Some notes:

- ValidatedSpec is basically a Spec with associated Results. As such, the Spec Model was modified to include a dictionary of results per Property. Thus, ValidatedSpec does not have a separate model, nor a separate Artifact Type, since it is simply a Spec with the results sections for each property filled in, and with a slightly different identifier. ValidatedSpec has always been a superset of Spec. However, if it is believed this can cause issues, we can discuss it and modify this as needed.
- The modules of the spec and validation packages were moved around a bit to better match the respective packages they are in.
- Since for now we assume that there is only one Spec per ns/model/version, Spec was modified to always use a default identifier "spec". The load() method was overloaded to use it, but not the load_with() method, as this would require giving default values for the store and context params. We can discuss this if needed as well.
- The id for a ValidatedSpec is "<associated_spec_id>.validated".


mlte/spec/spec.py
```diff
@@ -42,15 +43,15 @@ class Spec(Artifact):
...
-    ):
+    def __init__(self, properties: Dict[Property, Dict[str, Condition]]):
         """
         Initialize a Spec instance.
 
         :param properties: The collection of properties that compose the spec, with their conditions keyed by measurement id.
         :type properties: List[Property]
         """
+        # Standard identifier. Good enough?
+        identifier = DEFAULT_SPEC_ID
```
turingcompl33t, line 54: Thank you for pointing out this update in PR comments @sebastian-echeverria. Here are my thoughts on this:
- I think in the typical use case, there will only ever be one spec for each (namespace, model, version). The same is true for some other artifacts: negotiation card, validated spec, and report
- However, I would like to be able to support multiple artifacts of each type in the event that they are needed
- I'd like to handle this by doing something similar to what you have here; the four types of artifacts that I mentioned above should accept an `identifier` argument on construction, like any other artifact that has multiple instances (e.g. `Value`)
- However, we also provide a default argument, similar to what you have here, that is `default` followed by the artifact type
- For instance, for `Spec`, it would be `default.spec`

Let me know what you think of this proposal

mlte/spec/spec.py
```diff
@@ -124,6 +112,11 @@ def _to_property_model(self, property: Property) -> PropertyModel:
         }
         return property_model
 
+    @classmethod
+    def load(cls, identifier: str = DEFAULT_SPEC_ID) -> Artifact:
```
turingcompl33t, line 116: This should still fit in with the proposal above.

mlte/validation/validated_spec.py
```diff
@@ -0,0 +1,111 @@
...
+        """
+        Generates a model representation of the ValidatedSpec.
+
+        :return: The serialized model
+        """
+        model = self.spec.to_model()
+        model.header.identifier = f"{model.header.identifier}.validated"
+
+        # Add results to main model.
+        spec_model: SpecModel = cast(SpecModel, model.body)
```
turingcompl33t, line 62: This is the only other thing I am unsure about. If we reuse the `Spec` artifact type in the `ValidatedSpec` model, won't it be hard to distinguish between these two types, for instance in the frontend? Would it be difficult to simply reuse the model components from `SpecModel` in a `ValidatedSpecModel` type that has its own artifact type?

turingcompl33t (changes requested):
Two comments on design. Let me know if you thing these are wrong / difficult to change.

mlte/spec/spec.py
sebastian-echeverria: I agree with the default idea for most artifacts, but allowing for different ones if needed. I agree about the default.artifactype for the default value, sure. Identifier is the first param in the artifact constructors, so if it has a default value as an argument, all other arguments must have a default value as well. I can add this for Spec and ValidatedSpec.

mlte/spec/spec.py
sebastian-echeverria: Yes, for load(). Do note, however, that load_with has 3 params: id, context, and store. If we provide a default value for the id argument, what happens with context and store?

mlte/validation/validated_spec.py
sebastian-echeverria: Yes, it would be difficult. I was unsure about this structure too. If there are benefits, like the one you mention, I can create a simple ValidatedSpecModel that wraps or is similar to the Spec one to ensure the artifacts are treated differently, sure. I'll work on this.

mlte/spec/spec.py
sebastian-echeverria: Btw, one way to handle this would simply to force all of them to be named parameters, which would allow us to have a default value only for identifier.

sebastian-echeverria:
Ok, design suggestions applied to branch: 

- ValidatedSpecModel was created and a specific artifact type was created, reusing parts of SpecModel, and more cleanly separating submodels and tests.
- Default id was added to Spec, ValidatedSpec and NegotiationCard, and load() and load_with() slightly modified to allow using default id when needed.
- Also implemented #240, timestamp is in ArtifactModel.header and stored during save_with().
- Created common function in Artifact to build the ArtifactHeaderModel, to reduce code duplication.

mlte/artifact/artifact.py
```diff
@@ -88,37 +93,49 @@ def save_with(
...
+        cls, identifier: Optional[str] = None, *, context: Context, store: Store
     ) -> Artifact:
         """
         Load an artifact with the given context and store configuration.
         :param identifier: The identifier for the artifact
         :param context: The context from which to load the artifact
         :param store: The store from which to load the artifact
         """
+        if identifier is None:
+            identifier = cls.get_default_id()
```
turingcompl33t, line 137: This is wonderful.

turingcompl33t (approved):
Thanks for all of this great work @sebastian-echeverria. LGTM.

PR #245, opened 2023-08-28
Add CORS Configuration Option to Artifact Store

### LINKED ISSUE #244: Artifact Store Should Support CORS Configuration
@sei-aderr ran into the issue where CORS must be enabled for the frontend to make requests to the artifact store. We should add a configuration option to the artifact store that enables this.

turingcompl33t:
- Resolves #244 

This PR adds a configuration option to the artifact store to accept allowed CORS origins. The configuration is specified via the commandline:

```bash
mlte store --allowed-origins http://localhost:9000
```

or via the environment:

```bash
ALLOWED_ORIGINS='["http://localhost:9000"]' mlte store
```

### DIFF
--- a/mlte/cli/cli.py
+++ b/mlte/cli/cli.py
@@ -6,6 +6,7 @@ Top-level command line interface.
 
 import argparse
 import sys
+import traceback
 
 import mlte.frontend as frontend
 import mlte.web.store.main as server
@@ -96,6 +97,9 @@ def run():
     except KeyError:
         parser.print_help()
         return EXIT_SUCCESS
+    except Exception:
+        traceback.print_exc()
+        return EXIT_FAILURE
 
 
 if __name__ == "__main__":
--- a/mlte/cli/cli.py
+++ b/mlte/cli/cli.py
@@ -59,8 +59,14 @@ def _attach_store(
     parser.add_argument(
         "--backend-uri",
         type=str,
-        required=True,
-        help="The URI for the backend store.",
+        default=settings.BACKEND_URI,
+        help=f"The URI for the backend store (default: {settings.BACKEND_URI}).",
+    )
+    parser.add_argument(
+        "--allowed-origins",
+        nargs="*",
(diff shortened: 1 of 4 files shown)


turingcompl33t:
@sei-aderr what is the server side response when you get this response? Is it a 500? Or some other error code? Are there any other logs that are produced that you can provide? I can't reproduce this on my end because I don't have the frontend implementation on my side.

sei-aderr:
On the server side I am getting a 400 on an HTTP Options.

```
INFO:     ::1:57620 - "OPTIONS /api/namespace HTTP/1.1" 400 Bad Request
```

For testing you should just be able to run this in the javascript console in your browser as it looks to give the same response as when it runs within the frontend. The only difference being the origin will be whatever website you are on, and not localhost:8000.

```
fetch('http://localhost:8080/api/namespace', {
  method: 'POST',
  body: JSON.stringify({
    identifier: 'foo',
  }),
  headers: {
    'Content-type': 'application/json; charset=UTF-8'
  }
})
.then(res => res.json())
.then(console.log)
```

So for example when I run that when the window is on stack overflow,
image

With the console command on the frontend at localhost:8000,
image

And then through the actual code in the page,
image

sei-aderr:
And each one gets the same 400 on an options endpoint in the backend,
image

With this being the command I used to run the store, just in the demo store
```
mlte store --backend-uri fs://store --allowed-origins http://localhost:8000
```

turingcompl33t:
OK @sei-aderr, after some debugging time, it appears that the order in which middleware is applied and routes are injected matters. I reordered the two such that middleware is applied after routes are injected, and now the example above works.

When I execute this when the browser is on StackOverflow, the store should be run with:

```bash
mlte store --allowed-origins https://stackoverflow.com
```

not

```bash
mlte store --allowed-origins http://localhost:8000
```

because the origin is automatically set by the browser to `https://stackoverflow.com`.

sei-aderr (approved):

Issue #186, opened 2023-08-10
Implement UI Integration with Store: Versions


turingcompl33t:
The MLTE UI should be capable of performing version operations. These include:

- Create a version
- Read all existing versions and display them
- Delete a version

Endpoints for all of these operations are supported by the current implementation of the MLTE artifact store.

turingcompl33t:
Resolved by #249

Issue #185, opened 2023-08-10
Integrate UI Integration with Store: Models


turingcompl33t:
The MLTE UI should be capable of performing model operations. These include:

- Create a model
- Read all existing models and display them
- Delete a model

Endpoints for all of these operations are supported by the current implementation of the MLTE artifact store.

turingcompl33t:
Resolved by #249

Issue #184, opened 2023-08-10
Implement UI Integration with Store: Namespaces


turingcompl33t:
The MLTE UI should be capable of performing namespace operations. These include:

- Creating a new namespace
- Reading all existing namespaces and displaying them
- Deleting a namespace

Endpoints for all of these operations are supported by the current implementation of the MLTE artifact store.

turingcompl33t:
Resolved by #249

PR #249, opened 2023-09-05
Initial API Endpoint Integration

sei-aderr:
Addresses #184, partially addresses #185, #186. Able to read and display versions + models but not add new ones through the UI

### DIFF
--- a/mlte/frontend/nuxt-app/pages/index.vue
+++ b/mlte/frontend/nuxt-app/pages/index.vue
@@ -4,7 +4,7 @@
       <div style="padding-top: 115px">
         Namespaces
         <hr />
-        <div v-for="namespaceName in namespaces" :key="namespaceName">
+        <div v-for="namespaceName in namespaceOptions" :key="namespaceName">
           <SidebarItem
             @select="selectNamespace(namespaceName)"
             @delete="deleteNamespace(namespaceName)"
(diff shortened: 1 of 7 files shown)


mlte/frontend/nuxt-app/pages/index.vue
```diff
@@ -158,25 +163,21 @@ const path = ref([
   },
 ]);
 
-const namespaces = ref(["Default", "TEST 1", "Super longg purposes"]);
-
-const selectedNamespace = ref("Default");
+const { data: namespaceOptions } = await useFetch<string[]>(
+  "http://localhost:8080/api/namespace",
```
turingcompl33t, line 167: Not necessary to update this for this PR, but we will need to make this configurable to alllow users to run the store in locations other than `localhost:8080`. We can default to some value like this in development mode, but when run from the CLI, we should accept an argument that is the address of the artifact store.

turingcompl33t (approved):
Nice work Alex. Our first glimmers of an end-to-end MLTE system!

mlte/frontend/nuxt-app/pages/index.vue
turingcompl33t: I created an issue for this.

sei-aderr: Yes, still want to get that implemented. I am not sure that there is going to be a simple way to make that happen so I just wanted to get this along with the rest of the frontend implemented before going down that rabbit hole.

Issue #253, opened 2023-09-07
Remove package-lock.json From Checkin


turingcompl33t:
I had issue when running the frontend locally for development because Alex had resolved packages from an SEI internal package repository. This is fine, but in order to make sure this doesn't happen, we can just remove the package-lock.json file from the checkin. This will force all developers to resolve dependencies fresh, but for now this is OK.

turingcompl33t:
Resolved by #254.

Issue #243, opened 2023-08-23
Frontend Parsing Layer


turingcompl33t:
We need some mechanism to parse the JSON representation of MLTE artifacts to in-memory data structure representations. This will enable the frontend code to be logically separated from the object parsing functionality. These representations can be "static" in the sense that they will never be modified by complex logic on the frontend. The content may change (for instance, when updating content for writeback to the store) but these modifications will take place by accessing the fields of the object directly.

For non-polymorphic artifacts, there is a single structure that is the parse target. This will essentially equate to re-implementing the "model" representation from the MLTE backend in the frontend.

For polymorphic artifacts, like `Value`, there may be multiple parse targets. For instance, `Integer` looks different from `Real`. However, for the purposes of the frontend, it may be sufficient to parse to a generic type (just `Value`) and leave further parsing to a later implementation. At the very least, this will allow the bare minimum functionality to display information about individual artifacts in the frontend store UI.

turingcompl33t:
Resolved by #254.

PR #254, opened 2023-09-07
Feature/frontend artifact layer

sebastian-echeverria:
Addresses #243 and #253.

Notes:

- We decided not to add an explicit artifact type layer, as the type of restrictions we wanted to imposed for this are now well supported by Javascript, and barely for static typing for TypeScript. To add runtype checks, there were some potential libraries to help out, but creating the types for this was a convoluted process, which would add a lot of complexity for very little benefit. Adding an explicit layer would also add more maintenance work, as it would need to be synched with the Python version; while this may be worth it if it worked as intended and supported the type features that Javascript doesn't, the current negatives of all of this outweigh the positives. Maybe in the future, if more complex logic is added in the frontend, this could be explored further.
- Instead, functions to validate JSON objects in Javascript/TypeScript in the frontend were added. This would allow the validation of JSON objects received by the frontend, to check if they comply with the proper JSON schemas for artifacts of that type. Note, however, that only the functions were added, and this is not integrated into the actual code receiving artifacts, so a bit more tweaking may be needed when doing this integration. The integration, however, should be relatively straightforward: call the validation when JSON objects are received.
- In order to have the artifact JSON schemas available to the frontend, we modified the schema Python tool to generate the schemas in two locations: the original one, and a copy inside the frontend folders. We also included this copy as files in the repo. Even though this means that the schema will be duplicated in the repo, both copies are manually update by the Makefile commands, so this doesn't add more work, only potentially some minor confusion. Another option was to use gitignore and not add the copy of the schemas, but still generate them by the tool; however, this would require an additional setup step to people working with the repo, of manually calling the schema generation tool just for the copies to show up in the frontend folders, which doesn't seem to have much of a benefit and would complicate env setup.


mlte/frontend/nuxt-app/composables/artifact-validation.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export function isValidValue(artifact: object) : boolean {
+    return isValidArtifact(artifact, valueSchemaData);
+}
+
+// Validates if a given JSON string is a valid JSON artifact for the provided schema.
+function isValidArtifact(artifact: object, schema: object) : boolean {
+    if(!("body" in artifact)) {
+        console.log("INVALID!!! Object has no body section.")
```
sebastian-echeverria, line 30: Some of these printouts were left in for now to help debugging when integrating, but they can be removed later.

turingcompl33t: Good call.

turingcompl33t (approved):
LGTM @sebastian-echeverria. I think this is a great, low-overhead (effort-wise) solution.

Issue #233, opened 2023-08-22
`CPUStatistics` and `MemoryStatistics` Should use New `Value` Implementation


turingcompl33t:
@sebastian-echeverria pointed out that these types do not inherit from the new `Value` hierarchy.

turingcompl33t:
Resolved by #255.

PR #255, opened 2023-09-08
Fixed CPUStatistics and MemoryStatistics, demo, and schema tests

sebastian-echeverria:
Mostly addresses #233, but also made some other fixes:

- Addresses #233 by making CPUStatistics and MemoryStatistics derive from ValueBase,  as well as fixed and re-enabled unit tests for these value types.
- Fixed and re-enabled schema unit tests for Spec, ValidatedSpec and Value schemas.
- Fixed demo, ensuring evidence and report notebooks work properly with new version of objects (except for report generation), and fixed bug in ConfusionMatrix which was not properly serializing a numpy array.

### DIFF
--- a/demo/confusion_matrix.py
+++ b/demo/confusion_matrix.py
@@ -6,9 +6,10 @@ Implementation of ConfusionMatrix value.
 
 from __future__ import annotations
 
-from typing import Any, Dict, List
+from typing import Any, Dict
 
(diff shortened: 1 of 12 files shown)


turingcompl33t (approved):
LGTM. Thanks @sebastian-echeverria!

PR #256, opened 2023-09-10
Remove `jsonschema` As a Runtime MLTE Dependency

### LINKED ISSUE #200: Remove Dependency on JSON Schema
We used to use the `jsonschema` library for schema validation. However, with the migration to `pydantic`, this should no longer be necessary. We should remove this dependency.

turingcompl33t:
- Resolves #200

We no longer rely on `jsonschema` as a runtime dependency for schema validation; this is handled by Pydantic. This dependency is still included, but only for unit test execution.


sebastian-echeverria (approved):
LGTM!

PR #263, opened 2023-09-25
Modified Property so it can load properties from any module

### LINKED ISSUE #262: Allow extending new Properties outside of MLTE packages
Due to the way Properties are loaded from a Spec when a Spec is loaded from a Store, Properties can only be located inside the mlte.property package. However, while developing the Quality Attribute Scenarios demo, we realized there are many Properties that may be needed when using MLTE that do not come with it, and sending a PR to add them is not a very flexible way of handling this. Even though it makes sense to eventually add common Properties to the main MLTE library, it also makes sense to allow people to extend it by adding new Properties that can be located anywhere, outside of the MLTE packages.

sebastian-echeverria:
* Resolves #262 by storing the module of a Property in a store, allowing to load Properties from any module thus stored.
 * Also modified built-in Properties to not be loaded by parent package, but only directly, to be consistent with this change in other packages.


turingcompl33t (approved):
Thanks for improving this @sebastian-echeverria, this has always been a rough edge for `mlte`. Also, thank you for removing the redundant type annotations when you come across them! LGTM.

Issue #265, opened 2023-09-28
Property Base Type Should be in base.py


turingcompl33t:
Rename `property/property.py` to `property/base.py` to be consistent with other sub-packages.

sebastian-echeverria:
Closed by #267

PR #267, opened 2023-10-03
Feature/simpler properties

sebastian-echeverria:
- Addresses #265 
 - Slight refactoring of base Property class to make it easier to create new Properties. Now, instead of people having to manually pass self.__class__name and __name__ as parameters to the base Property constructor, just an instance of the derived property is passed (along with the description and rationale), and the base Property constructor obtains those two items. This makes extending for new Properties much more user friendly.

### DIFF
--- a/mlte/property/base.py
+++ b/mlte/property/base.py
@@ -1,5 +1,5 @@
 """
-mlte/property/property.py
+mlte/property/base.py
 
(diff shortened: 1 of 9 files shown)


turingcompl33t (approved):
LGTM.

PR #270, opened 2023-10-04
Refactor Enum Serialization

turingcompl33t:
This PR refactors enum serialization for MLTE artifacts.

Previously, I had used `auto` to automatically generate the value of the `enum`. However, without the `StrEnum` type, this feature is only capable of generating integers for the `enum` value, making it difficult to recover the `enum` from the serialized representation.

To fix this, I just removed use of `auto` and inserted explicit strings manually for the enumerations.


sebastian-echeverria (approved):
LGTM. The only question I have, which does not prevent this from being merged, is the Config subclasses, with use_enum_values = True. I don't see this being used. What is its purpose?

PR #272, opened 2023-10-04
Fixes missing Spec id in stored ValidatedSpec

### LINKED ISSUE #271: Spec Identifier not being saved in ValidatedSpec
The ValidatedSpec has a spec_identifier value in its model, where the id of the spec this was validated from should be. However, it is apparently not being saved. Fix it.

sebastian-echeverria:
- Fixes #271, spec id not being stored for validated spec, and equalities that should have checked that as well.


turingcompl33t (approved):
LGTM.

Issue #264, opened 2023-09-26
Create Array Value type


sebastian-echeverria:
A Value type of that implements an Array would be very useful as part of the base MLTE package. While using MLTE to create different conditions, we ended up creating an Array Value, and all other Values just derived from this one. Deriving from it also makes it easier to create new conditions, as only the validation methods have to be added for simple Values that can be stored as an array. There is an example in the feature/new-properties branch, demo-2/values/array_value.py

krmaffey:
Partially addressed by PR #[306](https://github.com/mlte-team/mlte/pull/306), but ongoing conversation was moved to issue #323

sebastian-echeverria:
Closing to avoiding having it duplicated. Work to be continued in #323

Issue #269, opened 2023-10-04
Refactor Report in Demonstration


turingcompl33t:
The `Report` construction in the demo is broken because it does not reflect the new artifact structure. Fix the demonstration to build the report properly.

PR #273, opened 2023-10-04
QAS Demo - Flower Model

sebastian-echeverria:
This PR adds a second demo to the demo-2 folder with a flower model. Some notes and comments about it:

- This demo is fully self-contained, so even if it has issues, it should not affect the rest of MLTE (low risk of merging PR, even if not fully reviewed)
- The main goal of this demo is to show a use case where we start from a list of Quality Attribute Scenarios, and get to create a Spec, gather that data, and validate it in MLTE. 
- There is some how-to guidance being developed to guide through the demo (mostly on how to extend MLTE), but as it is the case with the original demo, it should be understandable in itself, and by default this guidance won't be inside the demo itself, so it shouldn't be a deterrent to move forward.
- A secondary goal of this demo was to start with some real Jupyter Notebooks, and translate them to using MLTE, to find challenges in real use of MLTE. Some of these challenges generated changes already incorporated into MLTE, some are pending issues with suggestions of improvement.
- In practice, related to the above, this demo makes use of adding new Properties outside of MLTE to add 4 new Properties needed for the QAS that are being validate. Several of these Properties should be merged into MLTE itself at some point, but for now having them isolated is simpler.
- This demo also creates an Array value, which was very useful for most other Values that needed to be created to add validation methods for the different Conditions. Issue #264 was created to evaluate integrating it into MLTE.
- This demo contains several small data files (csv, a json model, and one image) that are used through the demo. None of them are big, but if we are not sure if we want to include all of them in the repo itself for different reasons, I can move them somewhere else and download them as needed (for example, not sure about the image permissions, or permissions for some of the CSV flower data). In particular, the actual model weights used for the demo is a huge 100 MB file that was instead uploaded to our CMU's Google Drive, and downloaded through a script during the demo execution.  We may want to figure out if that is the proper place for that or other files as well.
- There is a substantial overlap between this and the original demo (besides the overall structure). It may be beneficial at some point to merge both of them into one, but we can discuss this later.


sebastian-echeverria:
- The only thing still missing from this demo is the Report generation, in line with how it is missing in the default demo, as per #269

krmaffey (changes requested):
Sebastian, this demo is wonderful! It looks great to me, I think we are ready to merge it in. I discussed this with Kyle and he agrees, but we do have a request with regards to file structure.

Could you please change it so that this new demo is under the same folder as the old one? So the file structure will look something like this:

demo/
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;simple/
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;requirements.txt
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;requirements.ipynb
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;etc
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;scenario/
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;requirements.txt
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;requirements.ipynb
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;etc

I am not committed to those folder names (simple and scenario), if you have a better idea feel free to do that. My original thought was iris/ and flowers/ but I think it might be confusing since those are quite similar and the demos are distinct.

turingcompl33t:
Agree with @krmaffey, just want to change the names of the directories now that we have two demonstrations, and this will be good to go!

krmaffey (approved):
Looks great to me, thanks Sebastian!

PR #274, opened 2023-10-04
Changes memory  measurement to be cross-platform

### LINKED ISSUE #268: Add macOS support to memory profiler
Current memory profiler only works on Linux. Either modify or add another one that works in more platforms, especially in macOS.  Fox example, resource.getrusage could be used to extend this to macOS and BSD in general.

sebastian-echeverria:
- Fixes #268 , using psutil library instead of pmap to measure process memory. This is very useful when developing on macOS, for the demo to run transparently. This should support macOS, Linux and Windows, though it was only tested in macOS and Linux.


turingcompl33t (approved):
This is great, thanks for fixing this longstanding issue! LGTM.

PR #275, opened 2023-10-06
Demonstration Update

### LINKED ISSUE #269: Refactor Report in Demonstration

turingcompl33t:
This PR updates the demonstration to include a negotiation card and an updated MLTE report. Aside from being useful for demonstration purposes, this should make it easy to produce these artifacts for development purposes.

- Resolves #269


sebastian-echeverria (changes requested):
The demo update LGTM, and it also works for me. 

Regarding the integration between Report and ValidatedSpec, even though having the report being standalone has its advantages, I do like the idea of it referencing the ValidatedSpec by id to simplify things. In practice, the output of the report will either be the page on the UI or an exported PDF from there, so I do not think it is much of an issue for the report to be not self-contained underneath.

Regarding doing the same for ValidatedSpec and Spec, it is an interesting idea, but it currently wouldn't work. The way the model is build right now, each of the results is associated to each of the properties, so it is not currently just embedding the Spec in one location and the results in other. However, that being said, the in-memory artifact does have them separate, and we could modify the model to store them separately as well. It does make sense to avoid repeating the Spec information. However, in this case we would not only refer to the Spec id from the ValidatedSpec, the ValidatedSpec would also have to refer to each Property, since the results are associated by Property, so there would be some duplication in terms of the results being stored by Property. It could be done, but we should give it more thought, separately, to see the upsides and downsides, as it is not that straightforward as it would be with the Report.

sei-aderr:
Updated demo also worked for me to create the negotiation card and report beside the (expected) failures with memory items on mac

sebastian-echeverria:
Note that this version of the demo should not show memory failures on macOS. This branch already includes a fix from a previous PR that allows memory stuff to work on mac. So if it is not working, it may mean you have not compiled the version in this branch, or the demo is using a previous version and needs restarting.

turingcompl33t:
@sebastian-echeverria I updated this PR with the new method of implementation for "nested" artifacts - the `Report` now merely references the `ValidatedSpec` with which it is associated. I also introduced a rudimentary system for executing pre-save hooks for artifacts. This allows, for instance, the `Report` artifact that the identifier for its validated specification (if it is given) is valid at the time the artifact is saved.

Of course, this does not prevent one from deleting the validated specification out from under the report after it is written, but we haven't written any logic concerning artifact deletion yet, so that will have to be something for a later time.

sebastian-echeverria (approved):
LGTM!

PR #276, opened 2023-10-06
Artifact UI

sei-aderr:
Currently in progress PR to finish the integration of the UI with the artifact store. Will update with comments as tasks are completed, feel free to test and check things out as I go so that I can get as many questions answered and fixes addressed before I am out next week 

Remaining todo:
- ~Finish card loading~
- Finish card saving
  - Mostly complete. Although the backend does not allow you to update an artifact if it has the same name as one already in the store, essentially making the edit button in the negotiation card table lead you to a form you can never submit. When selecting start a new negotiation card, the artifact id will not be passed as a url parameter and there will be an input box at the top of the form allowing you to specify what ID you would like to save under
  - Testing note: If you try and submit the form with ML problem type, or any of the data classification selects set to the unselected `- Select -` options, the form will not submit. It will not pass schema validation as an empty string is not a valid option for those fields. Not sure it is possible to remove that default option but here is the documentation for the component (https://patrickcate.github.io/vue-uswds/?path=/story/components-usaselect--default-by-default-select)
- Report loading
- Report saving
- Add handling for missing url params in both save and load methods

Issues  that may not be finished in this PR and should be transitioned to issues
- Current implementation of main page only allows for a single version to be selected to allow the UI to know which model/version to pass when selecting new card or new report
  - For this single select implementation, the selectedModel and selectedVersion variables are used to maintain that data. Not the most robustly implemented as it should be a removed feature. If it stays around, it should be improved
-  Card and report tables are not sortable
   - The vue implementation of the USWDS I am using (https://github.com/patrickcate/vue-uswds) does not allow the flexibility to add buttons within the UsaTable elements. So I adjusted the card and report tables to use the raw html/css version straight from USWDS to add the edit buttons to each row of the tables. However similar to the accordion expands, using the raw version does not correctly create the sort buttons in the table. Even the copy pasted examples from their website do not work for me in the project (https://designsystem.digital.gov/components/table/#sortable-table-rows). My only idea is that the configuration of the USWDS in the project isn't correct and potentially not pulling in the javascript assets for the buttons and their functionality but won't have time to troubleshoot that before this PR is merged.


sei-aderr:
Addressed: Negotiation card loading

sei-aderr:
Addressed: Negotiation card saving to the extend that is possible at the moment

turingcompl33t (approved):
LGTM.

PR #278, opened 2023-10-07
Format Demonstration Notebooks

### LINKED ISSUE #277: Demonstration Notebooks Should Be Formatted
We should run the `black` formatter on our demonstration notebooks.

turingcompl33t:
This PR updates our `format` and `check-format` targets to include notebook files from the demonstration.

- Resolves #277  

Note that this PR is based on #275 and therefore should be merged after that one is.


sebastian-echeverria (changes requested):
Overall it looks good, I think the only change is on the Makefile, but hard to check since #275 has not been merged yet.

Regardless, I have a couple of comments related to the dependencies:

- This comment should have been in #275 but it also applies here: I see you added "extras" to the "black" dependencies, specifically the "jupyer" extra. I am not entirely sure how the extras are handled by poetry, but regardless, "jupyter" seems to be a very old and not updated version of jupyter notebooks (last updated 2015), while "notebook" is the current package that is constantly updated, which is the one I had been using. I would strongly recommend changing that dependency to "notebook".
- Again, not sure how the extras work, but I see you also added ipykernel as a dev dependency. Both ipykernel and notebook are required to run the demo, so why are they added in a separate way? This is probably not a problem, just me not understanding enough of poetry, but I'd appreciate an explanation.
- Third, besides notebook and ipykernel, there are several other requirements needed for the demo, as detailed in demo/requirements.txt (including numpy, pandas, scikit and matplotlib). If you are adding notebook and ipykernel to poetry's toml dep management file, shouldn't we add those others there as well? Not sure where to best add them due to my lack of familiarity with poetry, but it feels weird to add two dependencies and not the rest for the demo.

sebastian-echeverria (approved):
LGTM!

I am ok with restricting the demo dependencies to requirements.txt, and now that is consistent. But I was thinking, maybe we want to add a note about that in the demo? Not sure if we are ever telling demo users that they should install those requirements for the demo to work.

PR #288, opened 2023-10-13
UI Updates

### LINKED ISSUE #281: New Negotiation Card Button
Edit the text in the button that allows a user to create a new negotiation card to read "New".

### LINKED ISSUE #282: New Report Button
Edit the text in the button that allows a user to create a new report to say "New".

### LINKED ISSUE #285: Refactor Main Artifact Store UI to Single-Select
Currently, we have (broken, I believe, according to @sei-aderr) support for multi-select of models and model versions on the main artifact store page. This is likely a result of mockups that did not perfectly reflect the intent for this page. We should refactor these to be single-select boxes (dropdowns) such that users may only select a single model and model version at a time. This should reflect the logic that is currently implemented anyway (users can currently only select a single model and version) and updates the UI to reflect this.

### LINKED ISSUE #286: Change Report Render Button to "View"
Currently, we have an "Edit" button for report artifacts. Eventually, we will likely add support for report editing back in to the UI. For now, though, it is sufficient to be able to render a report and nothing else. Therefore, we should change the button to read "View" instead of "Edit".

turingcompl33t:
- Resolves #286 
- Resolves #285 
- Resolves #282 
- Resolves #281 

This PR refactors the implementation of the index page in the MLTE UI to remove support for multi-select of models and versions and add support for single-select alone. In addition, it makes some small UI tweaks, and applies linting and formatting to the frontend source.

### DIFF
--- a/mlte/frontend/README.md
+++ b/mlte/frontend/README.md
@@ -62,21 +62,3 @@ All typescript code takes advantage of static typing. This type checking can be 
 ```bash
 $ npx vue-tsc
 ```
-
-```
-    <div class="split-div">
-    <b>Model(s)</b>
-    <div class="scrollable-div">
-        <div v-for="entry in modelOptions" :key="entry.model">
-        <ScrollableListItem
-            :selected="entry.selected"
-            @update="updateSelectedModels(entry)"
-            @delete="deleteModel(entry)"
-        >
-            {{ entry.model }}
-        </ScrollableListItem>
-        </div>
-    </div>
-    <br />
-    </div>
-```
--- a/mlte/frontend/README.md
+++ b/mlte/frontend/README.md
@@ -62,3 +62,21 @@ All typescript code takes advantage of static typing. This type checking can be 
 ```bash
 $ npx vue-tsc
 ```
+
+```
+    <div class="split-div">
+    <b>Model(s)</b>
+    <div class="scrollable-div">
+        <div v-for="entry in modelOptions" :key="entry.model">
+        <ScrollableListItem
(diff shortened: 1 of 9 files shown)


sei-aderr:
PR looks great! A nuxt developer in the making. Comments in no particular order,

- The TODO on line ~283 can be removed. If we are set on having 1 version, 1 model, you have it handled in the best way I am aware of
- Personal preference, but I would adjust the red paragraph tags, `Select a model and version...` to be after the `NuxtLink`s because you adjusted the New buttons to float left instead of right. You can also remove the TODOs about those as you implemented the model/version selector as you wanted
- The imports that got added like requestErrorAlert, and the ones in the composable Sebastian wrote aren't necessary. You can add global defines for them in the eslint if you want to do that to pass the linting. Or you can just leave them in there explicitly. The nuxt ones aka requestError, and responseError probably make sense to be ignored by eslinst, the composable I can see either way.

krmaffey (approved):
Thanks for your comments, Alex! Kyle, this looks good to me. Approving it so you can make changes based on Alex's comments as needed and then merge.

turingcompl33t:
Thanks for the comments @sei-aderr. 

- Removed the `TODO`s you mentioned
- Adjusted the paragraph tags; you are right, those looked terrible
- Added global defines for `requestErrorAlert` and `responseErrorAlert` to `.eslintrc.json`; I left the composables from Sebastian as explicit imports

krmaffey (approved):
LGTM, great work Kyle!!

PR #289, opened 2023-10-13
Report Rendering

### LINKED ISSUE #179: Integrate Report Rendering with MLTE UI
Currently, we generate a static MLTE report through a clunky server-side rendering implementation. This necessitates a remote (HTTP) request to an external server to generate the report, and was always something of a stopgap solution.

Now, we have a proper web UI. This web UI should be capable of rendering a MLTE report artifact in the same way that we were able to before with static report generation.

For this feature, we should add the capability that, when a report artifact is "clicked" in the MLTE UI, the report renders on the frontend. We should be able to repurpose some of the React code that we used to implement the original rendering functionality to the Vue frontend implementation that we have now.

turingcompl33t:
- Resolves #179

This PR adds the ability to render MLTE reports in the web UI.

### DIFF
--- a/mlte/frontend/nuxt-app/pages/index.vue
+++ b/mlte/frontend/nuxt-app/pages/index.vue
@@ -150,7 +150,7 @@
                         namespace: selectedNamespace,
                         model: report.model,
                         version: report.version,
-                        arfifactId: report.id,
+                        artifactId: report.id,
                       },
                     }"
                   >
@@ -167,7 +167,7 @@
                         namespace: selectedNamespace,
                         model: report.model,
                         version: report.version,
-                        arfifactId: report.id,
+                        artifactId: report.id,
                       },
                     }"
                   >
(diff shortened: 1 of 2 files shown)


sei-aderr:
This PR looks good to me too. It is not the final version so don't feel the need to nitpick it at all.

Everything loads into the page as expected for me. My one potential suggestion would be disabling each of the inputs so that the users can't edit the values in this temporary `View` mode, but not vital by any means

turingcompl33t:
@sei-aderr thanks for the review. I was trying to figure out how to disable the `UsaTextInput` and `UsaTextarea`, but I couldn't find the correct property to do so. Do you know what this is?

krmaffey (dismissed):
Kyle this is incredible. I don't know what property you need, but I can approve this PR!

sei-aderr:
You can disable the text input, and should work for any other input style field just like this. It looks like it doesn't adjust the color too much, just makes it non-selectable. If you want you can add some css for disabled elements to gray it out a bit more.
```
<UsaTextInput :model-value="form.summary.problem_type" disabled>
```

Issue #297, opened 2023-10-24
Integrate properties from scenarios demo into main mlte package


sebastian-echeverria:
The scenarios demo added several properties: Robustness, Fairness, and a couple related to cpu and memory cost when predicting/inferring (as opposed to when training). Move this from the demo and integrate it into the main mlte properties package, since they are generic enough that they should be useful to most people.

krmaffey:
Closed by PR#[306](https://github.com/mlte-team/mlte/pull/306)

PR #306, opened 2023-11-01
Feature/integrate demo

sebastian-echeverria:
Addresses #297 and #264.

More specifically:
- Moves 5 properties from demo/scenarios into the MLTE package itself.
- Converts the Array value from demo/scenarios into a full-fledged Value type in the MLTE package itself.

There may be some things to discuss in the Array value, like its functionality to extract subvalues.


mlte/value/types/array.py
```diff
@@ -0,0 +1,110 @@
...
+        """Comparison between Array values."""
+        return not self.__eq__(other)
+
+    def get_as_real(self, position: int) -> Real:
+        """
+        Return a value from the given position, as a Real value type.
+        :param position: The position to get the value from.
+        """
+        if position >= len(self.array):
+            raise RuntimeError(
```
turingcompl33t, line 88: Can we have this raise `IndexError` instead of `RuntimeError` to be consistent with Python's error handling for index-of-out-range conditions on builtin data types?

mlte/value/types/array.py
```diff
@@ -0,0 +1,110 @@
...
+        return_value.identifier = f"{return_value.identifier}.{position}"
+        return return_value
+
+    def get_as_integer(self, position: int) -> Integer:
+        """
+        Return a value from the given position, as an Integer value type.
+        :param position: The position to get the value from.
+        """
+        if position >= len(self.array):
+            raise RuntimeError(
```
turingcompl33t, line 103: Same here.

turingcompl33t (changes requested):
The only question I have about this PR is the "nature" of the `Array` value type. Specifically, I originally thought that `Array` would be a generic container for a sequence of other MLTE `Value`s. For instance, you could use it to implement a collection of `Integer` or `Real`, or you could use it to do something more unorthodox like a collection of `Opaque` or `Image` etc.

The implementation here is more specific in that the contained collection is represented as a "raw" Python list without semantics for the array elements. This is flexible and powerful, and likely something that users want, but my concerns arise when we consider issues of serialization and storage. This works now for Python `int` and `float` types because, in some sense, these types can be serialized automatically (by the API) and can analogously handled transparently by our current `Store` implementations (e.g. JSON knows how to handle a list of integers). However, thinking ahead to other `Store` implementations like RDBMS, it isn't clear how storage for this type might map onto that abstraction.

I suppose what I am working towards is
- Do we limit the contained type of `Array` to `int` and `float`?
- Do we create separate types for `IntegerArray` and `RealArray` that are specific to these two types, to remove some of the complications of `get_as_int` and `get_as_real`?
- What do you think of the "generic collection of `Value`? type? Do you think this is a `Value` implementation that might be useful? If so, we can maintain your current implementation here as `Array` (or something similar) and call that other one `List` or something like that.

mlte/value/types/array.py
sebastian-echeverria: Good catch, done

mlte/value/types/array.py
sebastian-echeverria: Same, done

sebastian-echeverria:
Very good questions about the nature of the array. Some thoughts:

- I think there is value in having a generic Array type that doesn't know about its contents. In practice, this is really equivalent to Opaque. Opaque is named that way because of how it is supposed to be used, but in reality, it is a Dict. And right now, Array and Opaque are the equivalents of Python's list and dict, but with the Value wrapper. I think your concerns about serialization into storages like relational databases make sense, and I had not thought about that, but I believe the same would also apply to Opaque, right? Not saying that is a reason not to address it, just saying it is something that should be addressed similarly for both cases.
- The get_as_real and get_as_integer are a bit clunky, and even though I thought in theory I was going to need them, in practice I didn't. That being said, if there is a need to have something like this, maybe extending Array into IntegerArray and RealArray would make sense, instead of having these very specific methods, and would be more extensible/maintainable. But I am still not entirely sure of a good use case where either option is needed, so that makes it hard for me to measure its importance of how to go about it. Not saying there aren't good use cases, just saying I have not thought of good ones yet. Now, notice that since currently get_as_integer and get_as_real work, IntegerArray and RealArray would basically have to be Arrays of Values, more specifically, Arrays of Integers and Arrays of Reals, and not Arrays of ints and Arrays of floats, so this bleeds into your next question. If we think instead of IntegerArray as an Array of ints and not of Integers, this wouldn't really help with the get_as_int stuff that much, and would rather only help to ensure we are checking that the types are the proper ones. (Yes, you could still have a get_as_integer method in an IntegerArray of ints that returns an Integer, but that is going back to use cases for this).
- There might be a good argument for having an Array of Values (which might be better calle ValueArray). But, again, I am not sure I see the use case yet. Since I have not had to play that much with models and such, it is not always easy for me to figure out good use cases for this type of thing. But if there is, that doesn't mean that an Array of primitives is not useful. Then, is an IntegerArray an extension of ValueArray, that has only Integer values? Or, as mentioned as another option above, an extension of Array, that only has int values?

Sorry for the long rambling. But to summarize my current thoughts (and feel free to write back):
 - I believe an Array (Pyhon list) type equivalent to the Opaque (Python dict) type is useful, while still being "opaque" in terms of the data it has. Opaque is more general due to the key/values it can store, but having a thin Array wrapper for Python lists seems useful, and any potential issues with it also apply to Opaque. (In fact, it is possible people will want to have a Value dict and they will use Opaque, even if it is not clear by name that it is so). So maybe we should have a separate ticket to deal with serialization to other storages for these Value types with "any" values in them.
 - To be consistent with Opaque, maybe we should just remove the get_as_integer and get_as_float, especially since I am not sure when they would be used. This should probably be part of thinking what we want to do with this: have a ValueArray of Values, and extensions for specific cases? Again, I think this might be useful, but it seems like a slightly different feature than an opaque Array. And I would like to think about use cases before making decisions on this.

turingcompl33t (approved):
LGTM. See #323.

PR #310, opened 2023-11-20
Feature/load all values

sebastian-echeverria:
Addresses #141 to allow loading all values from a given store with just one Value.load_all() call . More specifically

- Added value_class attribute to Value model, which stores full path to class that implements that value (string with module plus class name).
- Adds load_all_models() and load_all_models_with() methods to Artifact, to load all models of a given type in the given store.
- Adds load_all() and load_all_with() methods to Value to load all values from a given store. They load all models from Artifact, and then dynamically load each Value from its module name and class, to use from_model to load them into Values.
- Updated demos to showcase this new method.


turingcompl33t (approved):
These a great updates @sebastian-echeverria, thanks for taking the lead on implementing this feature we've wanted for a while. LGTM.

PR #333, opened 2024-01-05
Updating MLTE Mechanics

krmaffey:
Updated MLTE mechanics to explain how to run the front and back ends and get the namespace configured properly.


JennyNiemann (approved):
I think the instructions sound very clear!

Issue #338, opened 2024-01-18
Add Colors to Report Table


krmaffey:
In the old implementation of the MLTE report, results were colored as per their result (success=green, ignore=yellow, failure=red): 
![mlte_report_table](https://github.com/mlte-team/mlte/assets/49003949/c4767d6c-74b6-4522-9137-7dbf6e563fe2)
In the UI, the report table should have a similar coloring scheme to make it easier for users to communicate evaluation results.

PR #341, opened 2024-01-19
Feature/add report table colors

sei-aderr:
Addresses #338

### DIFF
--- a/docs/docs/development.md
+++ b/docs/docs/development.md
@@ -85,11 +85,20 @@ Unit tests failures result in build failures in CI.
 
 ## Front End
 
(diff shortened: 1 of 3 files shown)


sei-aderr:
@turingcompl33t Do you remember if accessing the status variable was working for you at one point? It causes errors when I run it, making the report table not load. It looks like it is a valid return of useFetch, but that may have been added in a later version of nuxt than what we are running for the project

sei-aderr:
Ready for merge pending review @krmaffey

krmaffey (approved):
Looks good, thanks Alex!

Issue #340, opened 2024-01-18
Default to Last Selected Model/Version in the UI


krmaffey:
In the UI, when a user leaves a report or negotiation card and returns to the main artifact store page, they have to reselect the model and version again to see their artifacts. This is clunky and wastes time, especially when there are several models and versions to select from. The artifact store page should remain on the model and version last selected when returning from a report or negotiation card.

PR #355, opened 2024-02-06
Add cookie for model and version selection

sei-aderr:
Addresses #340

Tested with a couple models and versions and everything was saving properly, did not mess with namespaces too much due to their pending removal.


sebastian-echeverria (commented):
Looks good to me, if it works with these changes. They seem minimal and use cookies, which makes sense as a way to maintain session data.

krmaffey (approved):
Super straightforward, this looks great Alex!

PR #356, opened 2024-02-12
Publish using MacOS

krmaffey:
Publish the package using MacOS; addresses issue with poetry build calling on a version of python that did not exist in the file system.
Updated the development docs to account for this issue.
New package version includes several new UI features!


sebastian-echeverria (commented):
Looks good, but I have a question, mostly just out of curiosity. Can't really replicate the process yet but I trust it works.

sebastian-echeverria (approved):
Ok, good enough for now, if it seems to work :)

Issue #361, opened 2024-02-26
Fix timestamp display in UI


sei-aderr:
Kate - Hi Alex! Jenny looked at the UI and I'm going back over some of her comments/ideas. One thing that we are both seeing is that some of the timestamps are still wonky - when I created a negotiation card in the UI, I get the right date/time. However, a specification I created in code is giving a timestamp from 1970. Here are some screenshots!

sebastian-echeverria:
Fixed by #363

PR #363, opened 2024-02-26
Fix timestamp display

sei-aderr:
Addresses #361 

Second attempt at fixing the timestamp display in the homepage table. All timestamps are now generated in the backend so they are consistently passed to the frontend in the same way and can be displayed correctly.


sebastian-echeverria (approved):
Seems to be working. I guess issue came from backend timestamps being in seconds, and frontend timestamps being in milliseconds, so there was no way to consistently convert them. Since now they are always in seconds, coming from the backend, this solution should work.

PR #364, opened 2024-02-29
Add Negotiation Card to Docs, Update README

krmaffey:
Adds the MLTE negotiation card into the documentation as a reference.

Updated the paper version from arXiv to IEEE in our README, and updated all README links so that they now open in another tab (which is our default for the documentation).

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,24 +11,28 @@
 
 ### Getting Started
 
-`mlte` is available on [PyPI](https://pypi.org/project/mlte-python/). Install the latest version with:
+`mlte` is available on <a href="https://pypi.org/project/mlte-python/" target="_blank">PyPI</a>. Install the latest version with:
 
 ```bash
 $ pip install mlte-python
 ```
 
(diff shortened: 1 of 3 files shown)


sebastian-echeverria (changes requested):
Overall it looks good, but for some reason, there are things that don't quite match the NegotiationCard model we have in the code. They are minor, but still. They include:

 - In the code, each goal has a description and a list of metrics, where each metric has a description and a baseline. In this doc it is not totally clear that each goal can have multiple metrics; the relation between goals and baseline/metrics is not as explicit.
 - In the Data, the code has an explicit Description field, which is not present here.
 - In the code, there is a Policies field, and IdentifiableInformation field. They seem to be the same thing here?
 - The code has an Input specification, besides the Output spec, but here I see only the Output one.
 - The Inference Requirements in the code are more detailed, containing exactly the same subfields as the ones described in the Development compute resources. Here it seems to be more free form.

Let me know if it doesn't matter, or if we should modify either the doc or the code to match better.

krmaffey:
Thanks for catching these discrepancies, Sebastian! I fixed most of them and actually just made the structure match that of the MLTE UI more closely. The only one that I think should be adjusted on the code side is the Identifiable Information field - I think that snuck out of the tooltip for Data Policies, because Data Policies covers things such as Personally Identifiable Information. I'll make an issue for that!

sebastian-echeverria (approved):
Looks good to me!

Issue #359, opened 2024-02-21
Refactor/Address tech debt in Negotiation Card and Report Pages


sei-aderr:
- [x] Simplify data model of report to directly load the backend object.
- [x] Load options for Problem Type and Classification drop downs from the schemas
- [x] Move `fetchArtifact` from report.vue into a composable
- [x] - Figure out the useFetch status bug, or just remove that check. Was an issue with nuxt versioning.
- [x] Adjust report findings to be stored outside of the form and have `loadFindings` populate that field correctly. 
- [x] Adjust useFetch within `fetchArtifact` to just use fetch
- [x] Improve how the version of the package is displayed in the frontend
- [x] Define types for findings in the report page, and any other missing types in the frontend
- [x] Adjust all calls in methods to not use useFetch, and instead properly use $fetch. Plan to be done by use of `fetchArtifact` method

sei-aderr:
@krmaffey @sebastian-echeverria 

Looking into package version display on the frontend. Currently I just had it as a hard coded value in the base layout, which has now become outdated. 

I see that it looks like the version is stored in `.bumpversion.cfg` and `pyproject.toml`. Not 100% what the best way to handle this would be. My simple but annoying solution would just be we adjust the version on the frontend layout manually whenever we update the version in those other files. Wondering if you guys had any other ideas

sei-aderr:
Loading the options from the schema is going to be backlogged/addressed in #362. If we go ahead with that idea then the schema loading is invalidated so not going to spend time implementing it now.

sei-aderr:
Some typing adding, some will be removed in namespace update so waiting for after that change to look into it again

sebastian-echeverria:
@sei-aderr 
> 
> Looking into package version display on the frontend. Currently I just had it as a hard coded value in the base layout, which has now become outdated.
> 
> I see that it looks like the version is stored in `.bumpversion.cfg` and `pyproject.toml`. Not 100% what the best way to handle this would be. My simple but annoying solution would just be we adjust the version on the frontend layout manually whenever we update the version in those other files. Wondering if you guys had any other ideas

We seem to be using a package called bump2version, which uses tha cfg file, to update versions as needed. If you look at the config file, you can see that when that tool is used, it updates the version not only in the .toml file, but also in one of the docs file. Apparently, it can update versions in different files as requested, as described in the docs/docs/development.md file. In fact, the version in the .toml file is not updated manually, but automatically when calling bump2version.

[...]

Issue #252, opened 2023-09-07
Make Artifact Store URL in Frontend Configurable


turingcompl33t:
Currently, the URL for the artifact store is hardcoded to a specific value in the frontend source. We should update this to make the value configurable.

In React applications I have written in the past, I determine this based on an environment variable that reflects whether the app is running in development or production. In development, we can continue to use a hardcoded value for the artifact store (say `http://localhost:8080`). In production, we must accept another parameter that contains the desired artifact store URL.

sebastian-echeverria:
Continued in #405

PR #367, opened 2024-03-07
Frontend refactors 

sei-aderr:
Addresses all points in #359


mlte/frontend/nuxt-app/pages/index.vue
```diff
@@ -305,11 +318,12 @@ async function selectModel(modelName: string, initialPageLoad: boolean) {
     selectedVersion.value = "";
   }
   if (modelName === "") {
+    versionOptions.value = [];
     clearArtifacts();
     return;
   }
 
-  await useFetch("http://localhost:8080/api/model/" + modelName + "/version", {
+  await $fetch("http://localhost:8080/api/model/" + modelName + "/version", {
```
sebastian-echeverria, line 326: Could you reduce the duplication of the hardcoded localhost:8080 path in some way? Such as by having a base fetch function that fetchArtifact and these other fetches call?

sebastian-echeverria (changes requested):
Changes look ok, but take a look at my comment below to see if we could add that as an internal improvement as well. Having that hardcoded server URL is a pain, but having it duplicated all over the code is worse. Might be a good time to reduce that duplication.

sebastian-echeverria (approved):
Nice! This actually addresses #252. Sending this to Kate for her review.

sei-aderr:
I think it is close to addressing that, but not 100% of the way there. Kyle's idea with that was to be able to feed in the backend URL with a CLI command or something so that it was configurable from a user perspective which one to point it at. I could set up a prod/dev endpoint or something similar but I am not sure how I would go about providing a way for the user to edit that value during the use of the package.

sebastian-echeverria:
Ok, fair enough. We can deal with that improvement later.

krmaffey (approved):
This looks great, thanks both!

Issue #335, opened 2024-01-05
Generate Elements Eagerly


krmaffey:
Right now, the only way to generate elements in MLTE (namespace, model, etc.) is to create artifacts that are children of the desired element, which then creates them. It would be useful to have the ability to do so eagerly. This could be accomplished by adding a parameter to set_context(), for instance, such as 'lazy=false' which would then create those elements if they do not already exist.

sebastian-echeverria:
General comment: strictly speaking, there is a way to create models/versions explicitly, though it may not be terribly obvious or convenient. ArtifactStoreSessions have methods to create both. The can either be access through the API, or directly. 

To do this through the API, it would probably need support for this in the Frontend, and I think that is not currently there.

For use inside code/notebook, creating model/versions can be achieved by:

```
from mlte.session.state import session
from mlte.context.model import ModelCreate, VersionCreate
session().store.session().create_model(ModelCreate(identifier="model_name"))
session().store.session().create_version(VersionCreate(identifier="version_id"))
```
Alternatively, this can also be done calling the create_parents function, like so:

```
from mlte.session.state import session
import mlte.store.artifact.util as storeutil
storeutil.create_parents(session().store.session(), "model_name", "version_id")
```

But I agree that neither of these options are efficient or user friendly, and require knowing details about MLTE that users may not need to know about, just to get this done.

Solutions include the lazy=false parameter to set_context, or could also be a simple, explicit call:

```
from mlte.session.state import session
session().create_context()
```

Or we could also add both options. Have a create method for the context in the session, and have a lazy param in set_context that calls that other method.

sei-aderr:
Yeah this is not currently available in any way through the frontend. At one there was the ability to make/delete namespaces which is now gone, and then the ability to delete models and versions which also lost its place.

I can look into that if we want to make it available from the UI

sebastian-echeverria:
Could you create an issue for that? We can discuss next meeting how important it would be.

krmaffey:
That makes sense, thanks Sebastian! The changes in your PR make sense for now and then we can discuss how important a frontend implementation would be.

sebastian-echeverria:
Fixed in #368 . Frontend potential UI in #373

PR #368, opened 2024-03-11
Eager Structural Parent Elements Creation

sebastian-echeverria:
Added optional functionality to eagerly create structural parent elements (model/version) for a context when using the MLTE library, either by calling a method or setting a param when setting the context. Addresses #335 .


krmaffey (approved):
LGTM thanks Sebastian!

Issue #375, opened 2024-03-15
Add authentication to frontend


sebastian-echeverria:
Add authentication to frontend. This implies:

- Adding login screen for user (simple user/pass for now, with OAuth2 form names)
- Obtaining token from /api/token endpoint when executing login
- Maintaining token in browser and using it when sending api requests
- Maybe redirecting to login screen when getting an Not Authenticated message

Details:
 - Token requests:
    - POST request should be made to the /api/token endpoint
    - Proper headers to include: "Content-Type": "application/x-www-form-urlencoded" and "Accept": "application/json"
    - Payload: a form with three fields: "grant_type": "password" , "username": [theusername], "password": [thepassword]
    - In JSON reply, token is a string contained in "access_token" field
    - There is also info about when the token expires, but we can kinda ignore that for now, for a bit at least
 - In every request to other endpoints, the token should be sent as a header in the following way:
    -  "Authorization": "Bearer [token]"
 - Not sending a token to an endpoint should get you a 401 Unauthorized result, which is when you should redirect to the login page.
 - Branch currently in development of this is, in sebastian-echeverria's fork, "feature/identity"

sebastian-echeverria:
Addressed by #388

Issue #374, opened 2024-03-15
Implement Authentication Mechanism


sebastian-echeverria:
Using selected framework, implement actual authentication checks.

sebastian-echeverria:
Addressed by #388

Issue #373, opened 2024-03-11
Add User CRUD to Frontend


sebastian-echeverria:
Add ability to see/modify users in frontend.

sebastian-echeverria:
Addressed by #388

Issue #372, opened 2024-03-11
Add User CRUD to API


sebastian-echeverria:
Wherever store the users will be saved to, CRUD will be needed to handle them. This requires the creation of functions to CRUD users.

sebastian-echeverria:
Addressed by #388

PR #388, opened 2024-04-05
Authentication

sebastian-echeverria:
Adds authentication to the backend API, as well as handling of this and user management from the frontend. More specifically:

 * Adds an OAuth 2 based authentication layer on backend, using Resource Owner Credentials (password) flow and tokens for access. Authentication server is embedded with backend, since no external integration is needed for now, to allow independence. Addresses:
   * #342
   * #369
   * #370
   * #374
   * #372
   * #375 
 * This implicitly adds support for a logically separate User Store to store and handle users. Currently, however, when the backend is started, the same store type/location is used for Artifact and User stores (#371 )
 * Access to the API now requires user and password for almost all endpoints, and as such access to frontend requires user and password as well. Default user is provided to get started.
 * Frontend provides page to manage users (#373 )
 * Addresses #238 as well


sebastian-echeverria (commented):
Other than the repeated code to fetch with tokens, and the improvements to error handling, both of which we said we would handle in the authorization epic, everything else on the frontend looks good to me.

mlte/backend/api/auth/authentication.py
```diff
@@ -0,0 +1,26 @@
...
+def authenticate_user(
+    username: str, password: str, user_store_session: UserStoreSession
+) -> bool:
+    """Validates the credentials."""
+    user = None
+    try:
+        user = user_store_session.read_user(username)
+    except Exception:
+        # Assume any exception means we couldn't load user it.
+        # print(f"Error reading user: {ex}")
```
sei-aderr, line 20: Did you want to leave this print and the one a couple lines down in as comments or remove/uncomment?

sebastian-echeverria: I wanted to leave them commented out intentionally, for now at least. They are useful when debugging, which we may need for the authorization part, but they leak some potentially sensitive authentication info if always enabled.

sei-aderr (approved):

Issue #423, opened 2024-06-20
Clear selectedModel cookie when model is deleted


sei-aderr:
Currently on the index page, if the selectedModel cookie does not match an option in the model list then the select will just show up as empty. Visually not much of an issue, but potential errors with get/puts with that value saved there and the new version input isn't disabled as it should be because it believes a model is selected.

PR #449, opened 2024-07-30
MLTE-423

sei-aderr:
Addresses #423 

Also fixes another bug when model names are just numbers


sebastian-echeverria (approved):
Looks good

Issue #464, opened 2024-08-28
Add negotiation card fields to report


sei-aderr:
Sub task of #463 to just add the remaining QAS and dev resource requirement fields to the report and re-order some others

sei-aderr:
Addressed in #466

Issue #460, opened 2024-08-23
Create additional form components


sei-aderr:
Form components for

- usage context
     - Tried a component that is in the repo. For whatever reason it would not load in the initial value with the way that I have it setup but would then update after

- Deployment platform + deployment mechanism
    - Probably will have the same issue as usage context, but will be an easy add if the issue can be figured out

sebastian-echeverria:
Was this addressed by #448 ?

sei-aderr:
It was not, was having some issues when I tried to do usage context so going to add it into the changes with the report evaluation.

sei-aderr:
Addressed in #466

Issue #463, opened 2024-08-26
Re-evaluate Report Form structure


sei-aderr:
"Take do a side by side look at the negotiation card, and the report form and see if it would make sense to mirror them as there are several inconsistencies. 

Some are easy and expected fixes, like ML Task in the negotiation card has helper text, where the field is Task in the report without helper text. But others like ordering of fields and content grouping under certain headers is different. So see if those groupings are still the way that you would like them to be or if they should line up more with each other"

Sub task - 
Add QAS and Dev Resources to report form as continuation from #459

sei-aderr:
@GALewis 

Please specify where you would like the QAS and dev resources fields form the negotiation card to be added in the report. Then follow what I mentioned above about looking for incostencies between the negotiation card and report forms and see if you want them to stay as-is or become more consistent

GALewis:
OK, I totally see what you mean about overlap. From a process perspective the report should not be editable. You should only be able to "View" or "Export". For View, the information should be more like a report and less like a form. I recommend the structure to be:
- System Information
- System Requirements - the scenario text for each
- Test Results - I know the structure exists somewhere
- Additional Context - The rest of the Sections of the Negotiation Card

Does this make sense?

sei-aderr:
I am open to the idea of most of those changes, but several decisions have to be made regarding them. 

1. While nearly all of the report could be imported from other sources, how would the comments field be handled? Currently the only way to add to it is through the form. There is also the quantitative analysis field that is not populated currently but I believe that would also be an import
2. What is the point of the view if the export exists? There is certainly the future possibility of there being fancy/interactive visualizations in the view that would require it to not be static, but that is currently not the case. So I don't see a reason to have two separate static views unless there is a clear reason to.
3. When you choose a negotiation card to import, would the data copy to the report and then not change or would it link back to the negotiation card so the data in the report could update as the card changes.

sei-aderr:
Also, would you view this as crucial to the "initial release" by the end of September? While this needs to be addressed I might backlog it in favor of the frontend work on the catalog store

GALewis:
The issue here is very similar to what we talked about this week. There is a disconnect between the new MLTE and the old MLTE and we never really revisited the Report. Let me try to answer your questions:

1. While nearly all of the report could be imported from other sources, how would the comments field be handled? Currently the only way to add to it is through the form. There is also the quantitative analysis field that is not populated currently but I believe that would also be an import

[...]

PR #466, opened 2024-08-29
Components for negotiation card sections and adding them to report

sei-aderr:
Addresses #460, #464

Each section of the negotiation card is now a component that is re-used between the card form and the report form. 

The negotiation card now includes all of the negotiation card fields as well as its extra couple.


GALewis:
Understood that the fields are still editable, for now. These comments are likely going to be for later then.

1. For System Requirements, we only need to display the scenario text (what is in the boxes).
2. The section that is called MLTE Evaluation should be Test Results (Quantitative Analysis)
3. For now, I would remove the Comments and Qualitative Analysis because this is going to change in the future. Once we have a better mapping between System Requirements and Test Code, we will be able to have Quantitative Analysis for those that have a mapping and Qualitative Analysis for those that do not. 
4. After Test Results, suggest to add a line and the text "Additional Context" to separate the rest of the information.

sei-aderr:
Addressed 2, 3, 4. Will add 1 as a comment on the #463 for when that one is fully addressed

sebastian-echeverria (approved):

Issue #471, opened 2024-09-06
Improve Documentation to Better Align with Current Process


GALewis:
The current documentation is very tool centric and should become process centric to provide better context for the tool and the artifacts supporting the process. We will center the documentation around the process with pointers to existing tool-centered content.

sebastian-echeverria:
Addressed in #482

PR #482, opened 2024-09-25
Initial Process Update to Documentation

krmaffey:
Initial updates to the documentation to make the MLTE process more central (#471). Looking for comments/feedback from @GALewis and then can adjust this (don't need to immediately merge it in).

Remaining TODOs (not an inclusive list):
* Add screenshots to the negotiation card step (3)
* Add screenshots to the test catalog step (6)
* Add in updated diagram with numbers instead of the placeholder diagram
* Look through the old MLTE framework page and pull over relevant info, then delete that page
* Parse through ‘using MLTE’ and determine if that is still needed and how to refine it if so


GALewis:
I am reviewing now. Will check in changes when I am done.

GALewis:
I pushed several updates to the branch. The only thing missing before the merge is the diagram for the schema in the Negotiation Card file. While it does need a lot of work, the Negotiation Card file does need to be included because it is linked from everywhere.

GALewis (approved):
I made several edits across all files. I am not sure if the Negotiation Card file is being included in the documentation, but I think it does with the minor changes that I made, even if there are not screenshots since it is linked from everywhere.

GALewis (approved):

Issue #486, opened 2024-09-26
Bug with readthedocs integration


sebastian-echeverria:
There is a bug with the readthedocs integration, since its yaml file is still pointing to the old requirements file that was removed from docs and merged into pyproject. Update yaml file to get dependencies from pyproject toml file.

sebastian-echeverria:
Closed by #487

Issue #485, opened 2024-09-26
Bug generating docs


sebastian-echeverria:
There is a bug related with backend/api/models that is preventing doc generation. A missing __init__.py file is breaking the doc build. Add it.

sebastian-echeverria:
Closed by #487

Issue #484, opened 2024-09-26
Integrate doc creation into CI process


sebastian-echeverria:
Integrate doc building into CI, failing CI if there are issues.

sebastian-echeverria:
Closed by #487

PR #487, opened 2024-09-26
Document Generation Fixes

sebastian-echeverria:
- Fixes bug that prevented doc generation (#485)
 - Addresses #484 by integrating doc generation in CI process, to detect and fix issues earlier in the process
 - Fixes issue with Readthedocs integration after dependencies were moved to pyproject file in Test Catalog PR (#486)


sebastian-echeverria:
Detailed review is not needed, this just fixes issues generating docs accidentally introduced in test catalog PR, and adds method to automatically detect them in the future.

sei-aderr (approved):

Issue #479, opened 2024-09-23
Minor Frontend Test catalog improvements


sebastian-echeverria:
A set of minor UI improvements: 

- The creation date of a catalog entry, which is generated/stored in unix time, is being shown as unix time, not as a time/date, which is the way it should be shown.
- There are at least 3 property categories that should be added to the list: fairness, interpretability, resilience
- The Code input field should be resizable horizontally as well as vertically
 - It would be useful to add a button that copies everything in the Code field to the clipboard, probably next to the Code name or so

sebastian-echeverria:
Addressed by #488

PR #488, opened 2024-09-26
Feature/minor frontend

sei-aderr:
Addresses #479


sebastian-echeverria (changes requested):
Please move the fairness, interpretability and resilience to the Property Category list. Also, if you can merge the current master into your branch, that would be useful to avoid the failing read the docs check. Thanks!

sei-aderr:
Addressed both

sebastian-echeverria (approved):
Looks good!

Issue #477, opened 2024-09-19
Extract test entries for sample catalog from current "scenarios" demo notebooks


sebastian-echeverria:
Go over Jupyter notebooks for the "scenarios" demo, and extracts parts of the code, especially from 1_evidence, splitting them by robustness/property. Export notebooks to Python code so we can include text and comments. Create .json files for each entry.

sebastian-echeverria:
Fixed in #489

Issue #473, opened 2024-09-09
Add sample test catalog examples from new scenarios to mlte repo


sebastian-echeverria:
Once the example notebook scenarios are done, follow similar steps to #4774 and add them as JSON files to the corresponding place in the mlte repo so they can be created when a sample catalog is set up.

sebastian-echeverria:
Fixed in #489

Issue #462, opened 2024-08-26
Create a default sample artifact store


sebastian-echeverria:
The Jupyter Notebook demos currently allow the creation of a sample artifact store, but they have no instructions on how to access it through the MLTE UI, and they may not be too easy to run. Find a way to have a sample artifact store and instructions to easily use it.

NOTE: it may be possible to improve the compatibility of the notebooks and have the instructions just be a script to run all those notebooks and generate the results. Note however, that this could be really slow, so having an updated version of this already store in the repo could make it easier for people to see how things look like when using the UI.

sebastian-echeverria:
Start at least with the negotiation card, in a subfolder in the /demo folder.

Update docs for getting started to point to this negotiation card.

GALewis:
Attached is a negotiation card for the flower identification model. Because it is simply an example, I want to make sure that it fits well with the model and the test code.

@rbedman Can you take the first pass and review everything, especially Development Compute Resources and the Input and Output Specifications. I also need help with the Data Section. The easiest, I think, would be to do this in the UI? The name of the model is flower-classifier and the version is 1.0

@sebastian-echeverria After Rob is done, your review would also be useful.

Thanks![flower-classifier-1-0.json](https://github.com/user-attachments/files/17155843/flower-classifier-1-0.json)

rbedman:
Grace,

I can't figure out how to view that in the ui - but I looked at the raw
javascript and think we are broadly in agreement for quality attribute
scenarios - slightly different wordings (mine are wordier - and somehow I
have two more).

Here is my version of the same that I made in the notebook. I've also got
the data in this version - something we should talk about on Tuesday is
that, given the number of labels, the data and label descriptions will be
harder to enter through the UI. We can discuss that on Tuesday

--Rob

GALewis:
Attached is a new version of the Negotiation Card. There are a couple of things to do:

@rbedman Please take a look again and make sure that data, input specification, and output specification are correct. Notice the values that I have marked as outputs that indicate problems to make sure that your code matches those values.

@sebastian-echeverria Can you do a sanity check, especially for resource numbers?

[flower-classifier-1-0.json](https://github.com/user-attachments/files/17192951/flower-classifier-1-0.json)

sebastian-echeverria:
In the wiki we have 3000 b as the storage for the prod resources, which is what we are using in the demo, so I would change that from the 128 GB it currently has.

GALewis:
[...]

Issue #444, opened 2024-07-29
Create sample scenarios to populate the test catalog


GALewis:
Create Jupyter notebooks for new sample scenarios, to be used as the base for creating entries int he test catalog.

sebastian-echeverria:
Fixed in #489

PR #489, opened 2024-10-04
Sample Catalog

sebastian-echeverria:
Addresses:
 - #444 
 - #473 
 - #477 
 - #462

More specifically:
 - Updates Scenarios demo notebooks to contain new QAS
 - Creates test entries for Sample catalog from Scenarios demo
 - Adds a sample negotiation card, and docs on how to use it with the UI


GALewis (approved):
Thanks for all the hard work!

Issue #546, opened 2024-11-26
Define MLTE Presentation Demo Environment


sei-secheverria:
Define the environment to use, as well as an easy way to setup the demo environment (frontend, backend, negotiation card, etc). This can be just a set of instructions, or a script if needed.

sei-secheverria:
Addressed by #549

sei-secheverria:
There is a new script in demo/scenarios, called run_environment.sh, that starts up a frontend and backend using the sample negotiation card and linking to the same store artifacts will be stored to when created from the Jupyter Notebooks in that folder.

PR #549, opened 2024-11-26
Demo/live demo 1

sei-secheverria:
Created script to easily start up environment for live demo. More specifically:

 - Added script that starts up backend and frontend, using sample negotiation card, and using same FS artifact store as the scenarios demo Jupyter Notebooks. This will simplify greatly starting up a live demo based on those scenarios. Addresses #546
 - Also made minor improvement to docker compose config file, allowing it to switch and use different env files for frontend and backend as needed, without having to modify default env files.


sei-aderr (approved):
Didn't test out all of the environment variable options but, just running `run_environment.sh` worked for me with default settings and brought up a frontend and backend with the default negotiation card present

sei-secheverria:
Thanks!

Issue #514, opened 2024-11-15
Convert Property to Quality Attribute Category


sei-aderr:
Current naming scheme of these MLTE concepts is confusing. Restructure them to make more sense and relate more directly to QAS we have made in demos. 

First task is to rename Property to be Quality Attribute Category

sei-aderr:
@sei-glewis @rbedman 

Are we all in agreement that we are going to move from "Property" and to instead use Quality Attribute Category to represent things such as cost, robustness, etc and Quality Attribute to represent the specific measurement such as storage cost, or robustness to blur

sei-glewis:
Yes. That is what I am using in the QA spreadsheet as well.

sei-aderr:
Addressed in #553

PR #553, opened 2024-12-10
MLTE-514

sei-aderr:
Addresses #514, renames property to Quality Attribute Category

Had some waffling between qa category, QA Category, QACategory in some of the docs. I tried to just follow what was there before with caps being the actual type, lowercase being the concept but open to feedback if we want to go more strictly towards one of them.


sei-secheverria (approved):
Looks godo to me

PR #555, opened 2024-12-18
Feature/demo improvements

sei-secheverria:
Minor demo improvements, including avoiding duplicating the negotiation card in the notebooks and using the sample one instead, ordering the names of the simple demo, and adding a general, basic script for a live demo using the notebooks.


sei-aderr (commented):
Pushed a typo fix where there was a double space. 

Having a small issue with the scenarios test script, when I run it from the mlte root directory, all the test run but it is unable to find `copy_nc.sh` but then the notebook test all run and work. If I run it within the `demo/scenarios` directory, it finds `copy_nc` but then all of the tests fail.

sei-secheverria:
> Pushed a typo fix where there was a double space.
> 
> Having a small issue with the scenarios test script, when I run it from the mlte root directory, all the test run but it is unable to find `copy_nc.sh` but then the notebook test all run and work. If I run it within the `demo/scenarios` directory, it finds `copy_nc.sh` but then all of the tests fail.

Good catch. Should be fixed now. Script is supposed to be run through "make demo-test" from the root folder, btw.

sei-aderr (approved):

Issue #561, opened 2025-01-08
Frontend crashing when loading negotiation card


sei-secheverria:
When loading a negotiation card to be listed, the frontend crashes and starts complaining about invalid URL that can't be built.

sei-secheverria:
Bug was caused due to change in one of the underlying Javascript libraries, probable jsonschema (which was updated from the 1.4.x version we were using to 1.5). Since we were not tracking package-lock.json, and our dependencies indicated libraries like this one could be upgraded, when building a docker container from scratch new versions would be installed, which apparently cause some issues. Fixed by adding package-lock with currently working versions to git and docker installation.

sei-secheverria:
Addressed in #562

Issue #560, opened 2025-01-07
Fix README issue in docker deployment


sei-secheverria:
Pyproject now uses the README.md, so the docker images now need that readme before poetry install successfully builds mlte. Add to both backend and frontend dockerfiles so the images can properly build.

sei-secheverria:
Addressed in #562

Issue #559, opened 2025-01-07
Scenario Demo fixes


sei-secheverria:
There are some issues with the demo when setting it up from scratch:
 - The second notebook assumes wget installed in the system using it. Change to curl for better compatibility.
 - Improve the requirements steps to clearly indicate Docker must be running, how to run notebooks, which dev doc steps to follow, etc
 - Ensure that overall script is clear and up to date

Issue #558, opened 2025-01-07
Consistency of Make commands


sei-secheverria:
The make commands are a bit inconsistent, with some fake check- commands that are not really needed, get gen/vet commands not following the same naming standard as the rest, and the qa/ci commands confusingly checking or changing things. Make everything consistent (checks start with check-, qa does all the fixes, ci only does ci checks).

sei-secheverria:
Addressed in #562

PR #562, opened 2025-01-08
Demo and versioning fixes

sei-secheverria:
Contains a series of improvements for the live demo, as well as fixes for versions that have changed and were not pinned.

 - Addresses #559 by updating demo readme to include more detailed informations on how to run demo, updating the dev doc to clarify and separate some setup steps, and using curl instead of wget in getting model for better compatibility
 - Addresses #557, by changing docs to remove use of removed "poetry shell" and explicitly creating venv, and adding workaround to package information to avoid "format" error
 - Addresses #561 by pinning JS frontend dependencies by adding package-lock.json to git and docker images
 - Addresses #560 by explicitly adding readme to docker image before building MLTE package (readme is now a dependency)
 - Addresses #558 by making make commands simpler and more consistent in naming and behavior
 - Addresses #563 by creating an intermediate docker image when building the frontend, to avoid leaving unused dependencies in the final image

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -22,7 +22,6 @@ tools
 build.sh
 LICENSE
 Makefile
-README.md
 
 mlte/frontend/nuxt-app/package-lock.json
 mlte/frontend/nuxt-app/.nuxt/
--- a/.dockerignore
+++ b/.dockerignore
@@ -23,7 +23,6 @@ build.sh
 LICENSE
 Makefile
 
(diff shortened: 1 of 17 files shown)


sei-glewis (dismissed):
Thank you!

sei-aderr (dismissed):

sei-glewis (approved):
Looks good.

Issue #505, opened 2024-11-13
Epic 1: Integrate Negotiation Card though QAS


sebastian-echeverria:
Integrate Negotiation Card to rest of the artifacts. This will be done through Quality Attribute Scenarios, in that the Negotiation Card will contain them, and they will be linked to the different TestCases.

sei-secheverria:
All subissues finished

PR #577, opened 2025-01-13
MLTE-552 - Custom List File System Store

sei-aderr:


mlte/store/custom_list/quality_attributes/__init__.py
sei-secheverria, line 1: A note for later: Grace will likely have to review the end result of these set of changes, not underneath but on the UI. One thing that is important to talk to Grace to is the limitation that , at least for now, one list will not automatically update depending on what is selected in the other list. How does the frontend work now? Are the two list totally unrelated in practice?

mlte/store/custom_list/store_session.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+# TODO : Have this inherit from ResourceMapper and figure out how to handle the mismatch in params
+class CustomListEntryMapper:
+    """An interface for mapping CRUD actions to custom list entries."""
+
+    def create(
+        self,
+        custom_list_name: CustomListName,
+        new_custom_list_entry: CustomListEntryModel,
```
sei-secheverria, line 40: For all these mapper methods, I may want to switch the places of these two arguments. At a high level, it makes sense to put the list and then the entry, but for consistency with the other mappers, which always have the entity first, it may be better to put that first, and then the "additional" info, in this case the list it belongs to.

mlte/store/custom_list/underlying/fs.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def _write_entry(self, entry: CustomListEntryModel) -> CustomListEntryModel:
+        """Writes a custom list entry to storage."""
+        self.storage.write_resource(entry.name, entry.model_dump())
+        return self._read_entry(entry.name)
+
+    def _set_base_path(self, custom_list_name: CustomListName) -> None:
+        self.storage.set_base_path(
+            Path(self.storage.sub_folder, custom_list_name)
+        )
```
sei-secheverria, line 129: I would add a comment her explaining what this _set_base_path is doing.  It is inconsistent/different than the other stores, and it is not obvious what it is doing, and why it is called in each method.

mlte/store/custom_list/initial_custom_lists.py
```diff
@@ -0,0 +1,95 @@
...
+                                try:
+                                    session.custom_list_entry_mapper.create(
+                                        CustomListName.QUALITY_ATTRIBUTES, entry
+                                    )
+                                except error.ErrorAlreadyExists:
+                                    # If default values are already there we dont want to overwrite any changes
+                                    pass
+                                num_attributes += 1
+            print(
+                f"Loaded {num_attributes} Quality Attributes for initial list"
```
sei-secheverria, line 92: This code is copy/paster from the lines just above. Except for the folder and list name, there is no reason to repeat this. Please move this out to a function that is called twice with the different params. This won't scale at all.

sei-secheverria: And btw, if this code is basically the same as the one used by the sample catalog, for this section, it would make more sense to extract it as an utility, to be used by both the sample catalog and this store.

test/store/custom_list/test_underlying.py
```diff
@@ -0,0 +1,81 @@
...
+        test_entry.description = new_description
+        _ = custom_list_store.custom_list_entry_mapper.edit(
+            test_list.name, test_entry
+        )
+        read_entry = custom_list_store.custom_list_entry_mapper.read(
+            test_list.name, test_entry.name
+        )
+        assert read_entry.description == new_description
+
+        # Test deleting a list.
```
sei-secheverria, line 74: Deleting an entry, I believe

sei-secheverria (changes requested):
Looks mostly ok, a couple minor things to improve.

mlte/store/custom_list/quality_attributes/__init__.py
sei-aderr: Yeah, I will have to talk with her when we get to the frontend portion. If we go the route of the QA being in specific QAC, assuming I receive it in the frontend in a nice format, it isn't hard to auto update the QA list when a QAC is selected.

The lists are totally separate in the frontend now, mainly because the QA are not actually a list, it is a free text box and the QAC are the a predefined list with drop down selection.

mlte/custom_list/custom_list_names.py
```diff
@@ -0,0 +1,12 @@
+"""mlte/custom_list/custom_list_names.py"""
```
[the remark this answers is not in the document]
sei-aderr: Pushed fix

mlte/store/custom_list/store_session.py
sei-aderr: Pushed fix

mlte/store/custom_list/underlying/fs.py
sei-aderr: Pushed fix

mlte/store/custom_list/initial_custom_lists.py
sei-aderr: Yeah, can definitely separate them out as a function for the two custom lists.

The custom list load is a bit different than the sample catalog after they open the file and handle loading it. I could add a switch in the util function off a param for which type of item you are loading but then probably has to be a magic string, or another enum. Could also do something like make a util for the loading of the file and pass back the loaded file, or the json loaded from it so that the the other methods can handle it from there, but in my opinion it is a bit of a strange piece of data to be passing back.

Any thoughts on what you think would be best?

mlte/store/custom_list/qa_categories/fairness.json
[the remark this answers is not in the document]
sei-aderr: It doesn't look like we have updated descriptions for this iteration of the list and Rob has requested that we wait until after the card sorting to see what they come up with from that.

I can make a task to address this comment after they have a better grouping if that is sufficient for now.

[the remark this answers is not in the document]
sei-aderr: Added #578 and linked it to the epic #505

mlte/store/custom_list/initial_custom_lists.py
sei-secheverria: Maybe the best option is to create an generator, let's say a get_json_resources(). This generator would get the path as a parameter and return the json.load() of each file. Something like this (haven't tested it):

```
# This funciton would end up as an util, I think in mlte/_private/reflection.py, where we have util methods that work on packages themselves):
def get_json_resources(package: Package) -> Generator[Any, None, None]:
        resources = importlib.resources.files(package)
        with importlib.resources.as_file(resources) as resources_path:
            with os.scandir(resources_path) as files:
                for file in files:
                    if file.is_file() and file.name.endswith("json"):
                        with open(file.path) as open_file:
                            yield json.load(open_file)

# To use it:
for json_data in get_json_resources(whatever_package):
    # Whatever each function does: create a catalog entry, etc.

```

sei-aderr: Made this update and implemented it for the catalog store as well. Ready for review

sei-secheverria (approved):

Issue #579, opened 2025-01-14
Add Support for Context Data to Store ResourceMappers


sei-secheverria:
ResourceMapper is a generic interface to guide the creation of parts of new stores, one for each resource type. However, it assumes only an id is required to access a record. In some cases, more context information is needed, since the ids may be unique to a sub group (model/version for artifacts, list for custom lists, etc). Modify the interface and its use so that it supports a more generic context. Then apply to custom lists in #576 as well.

sei-secheverria:
Addressed by #580

Issue #576, opened 2025-01-13
Look into inheriting ResourceMapper in CustomListEntryMapper


sei-aderr:
Resource mapper only expects single parameters for each of its methods as it is implied that you would know what "list" or group of resources is being added to, aka you know you are adding a user to the list of users. The `CustomListEntryMapper` is flexible because custom list entries are all the same between lists, so those methods in its implementation require the list name, and then the entry information. This breaks during CI because it is not correctly implementing the interface.

The band-aid solution was to remove the inheritance which works for all the currently used implementations of the `CustomEntryListMapper` but as more things need to use it, such as the API this will no longer be a working solution. Consider the best route forward that would work in all cases, and could also potentially improve how this is handled for the artifact store.

sei-secheverria:
Support for this is added for #579, included in PR #580. Once that PR is approved, it can be used to fix this. In practice, other than re-inheriting from ResourceMapper, nothing else SHOULD probably need changing in CustomListEntryMapper, as its CustomListName will be used as the "Any" context that now mappers can have.

sei-aderr:
Addressed in #591

PR #580, opened 2025-01-14
Feature/store improvements

sei-secheverria:
- Addresses #579 by adding a generic Any "context" param to all CRUD calls, that can be used for passing out additional data. This will allow #576 to be fixed.
 - Also merged Model with ModelCreate models, and Version with VersionCreate models, since they were not really needed as separate entities.

### DIFF
--- a/mlte/context/model.py
+++ b/mlte/context/model.py
@@ -9,29 +9,11 @@ from typing import List
 from mlte.model import BaseModel
 
 
-class VersionCreate(BaseModel):
-    """The model that defines the data necessary to create a MLTE version."""
-
-    identifier: str
-    """The identifier for the version."""
-
-
 class Version(BaseModel):
     """Model implementation for MLTE model version."""
 
     identifier: str
     """The identifier for the model version."""
-
-    # TODO(Kyle): In the future, we may implement new endpoints
-    # that allow one to GET /version to get all artifacts associated
-    # with a (model, version) duple
-
-
-class ModelCreate(BaseModel):
-    """The model that defines the data necessary to create a MLTE model."""
-
-    identifier: str
-    """The identifier for the model."""
 
 
 class Model(BaseModel):
@@ -40,5 +22,5 @@ class Model(BaseModel):
     identifier: str
     """The identifier for the model."""
 
-    versions: List[Version]
+    versions: List[Version] = []
     """A collection of the model versions."""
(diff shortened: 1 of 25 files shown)


mlte/context/model.py
```diff
@@ -9,36 +9,18 @@
...
-    """The identifier for the version."""
-
-
 class Version(BaseModel):
     """Model implementation for MLTE model version."""
 
     identifier: str
     """The identifier for the model version."""
 
-    # TODO(Kyle): In the future, we may implement new endpoints
```
sei-aderr, line 25: Was this implemented, or just no longer planned? I don't remember exactly but I believe we have something similar

mlte/store/base.py
```diff
@@ -174,70 +177,80 @@ class ResourceMapper:
     DEFAULT_LIST_LIMIT = 100
     """Default limit for lists."""
 
-    def create(self, new_resource: Any) -> Any:
+    @abstractmethod
+    def create(self, new_resource: Any, context: Any = None) -> Any:
```
sei-aderr, line 181: Does this work for the artifact store case where you are passing model and version? Would we have to pass them as a variable combined and then unpack that data when using it down the line?

sei-aderr (commented):
Looks good, just had a couple questions.

mlte/context/model.py
sei-secheverria: To be totally honest, I don't really get what Kyle meant by "we may implement new endpoints that allow one to GET /version to get all artifacts associated with a (model, version) duple". We have endpoints to get all artifacts for a model/version duple; currently, it is /api/model/{model_id}/version/{version_id}/artifact . Maybe this is an old TODO before this was done? In any case, it doesn't help for it to be in this file, as far as I can tell.

mlte/store/base.py
sei-secheverria: Good question. Yes, that is the idea. The idea is to have a generic "context" that provides as much info as the crud methods of a store require. In the case of the custom lists, it is simpler, as it can be replaced with just the list name. In the case of the artifact store, we would need to pack model and version somehow and then unpack them. Probably the easiest way would be to create something like a ModelVersion model that contains a model id and version id, and then inside the methods do things like model_version.model_id and model_version.version_id to use them, instead of how now they are using directly model_id and version_id.

sei-aderr (approved):

Issue #511, opened 2024-11-15
Restructure Spec into TestSpec


sebastian-echeverria:
Restructure the current Spec into TestSpec, moving out QAS/Properties information, and leaving only a list of TestCases (see #510 for that substructure). More specifically:

- Rename to TestSpec
- Have it contain a list of TestCases (#510)
- Have functions to measure() or validate() all TestCases, at least those that have enough info. More on #524 
- Update backend stores to work with this new structure
- Updated frontend as needed to work with this

sei-secheverria:
Addressed in #608

Issue #510, opened 2024-11-15
Create TestCase subartifact


sebastian-echeverria:
Create a new unit/subartifact, TestCase, which will be the base for TestSpec. More specifically, TestCases:

 - Will have some info from current Properties, including name and goal, but also a new QAS list of ids, pointing to existing QASs from the NegotiationCard
 - Will include a section for validation, see #508 for the structure that would go in there
 - Will include a measurement section, which will include as much info as possible about the measurement to be used to get data for this TestCase, including class, method, arguments, and output, automating as much about this as it can be done
 - Should allow people to create a TestCase without a measurement and validator, and attach them later
 - Should have a method to run the measurement, and another to run its validation
 - Should have a way to manually add data to the result when measurement can't be automated (is the Ignore type of result enough?)

TestCases could look something like this:

```
       "name": "accuracy across gardens",
        "goal": "check if model performs well across different populations",
        "qas": ["id_qas_1"],
        "measurement": {
            "class": "mlte.measurement.external_measurement.ExternalMeasurement",
            "arguments": ["name": "name of the thing", "data": "all data as a string"]
            "output_class": "mlte.value.types.real.Real",
            "function": "demo.scenarios.funcs.accuracy",
        },
        "validation": {
           ...
        }   
```

sei-secheverria:
Note that the suggested "validation" section would basically be the new Validator class from #508  , and Conditions can be removed at this point.

Issue #508, opened 2024-11-14
Create Condition/Validator without creating/extending a type


sebastian-echeverria:
Currently, adding a new Condition can be cumbersome, as it requires creating/extending an existing type and adding a function. This is in part because the build_condition method automatically extracts information from the method calling it, including method name, method arguments, and method class, and stores this along with the validator in the Condition. All this extra information is just stored for tracking purposes, and so that when someone looks at a report, they can see what the test is actually doing.

Even though in some cases, especially for built-in types, this may be desirable, we don't want to force people to create a new Value type just to define a validation/condition. Figure out how to handle this instead. One option is simply to ask the user to create a Condition instead of using build_condition, and passing the name and arguments manually, and allowing the class to be null. It also may be the case that a Condition per se is no longer needed, if all the data was was being shown here is basically serialized in the Validator inside, instead of being opaquely serialized as a pickle (see #507).  

Maybe the "arguments" (actually thresholds) can be passed to the Validator, as also described in #507, and thus do not need to be stored at the Condition level at the same time. In this case, the only contribution of Condition on top of a Validator is 1) a name, 2) a class that contained a function the Condition came from, which doesn't exist if a Validator is created manually. If that is the case, it may be better to eliminate the concept of Condition, and just have a Validator, that can be created either manually by a person, or by calling build_validator(); and in the second case, it automatically adds extra metadata, including the method name, method arguments/thresholds, and class it came from; in the first one, the method name and class would not exist, and the argument/thresholds would have to be passed manually. It may end up looking something like this:

Old Validator model:
```
 "predicting memory": {
                        "name": "average_consumption_less_than",
                        "arguments": [512000.0],
                        "callback": "gASVz...",
                        "value_class": "mlte.measurement.memory.local_process_memory_consumption.MemoryStatistics"
                    }
```

New Validator model:
```
{
        "bool_exp": "lambda value: sum(g >= threshold for g in value.array) == len(value.array)",
        "thresholds": [512000.0],
        "success": "All accuracies are equal to or over threshold 512000.0",
        "failure": "One or more accuracies are below threshold 512000.0",
        "callback_bool_exp": "aDsada...",
        "from_class": "mlte.measurement.memory.local_process_memory_consumption.MemoryStatistics",
        "from_method: "average_consumption_less_than"
}
```

New Validator when there is no value type this came from:
```
{
        "bool_exp": "lambda value: sum(g >= threshold for g in value.array) == len(value.array)",
[...]

PR #585, opened 2025-01-24
Validator Refractoring, Part 2

sei-secheverria:
Addresses most of #508, improving Validators. More specifically:

 - Creates a ClassMethodInfo class that extracts functionality (previously hidden) inside Condition, to gather information about the class, method and arguments of the caller to current caller to a specific function.
 - Modifies Condition to use such class.
 - Modifies Validator to use such class as well, inside a new build_validator method. Validator now will have all information that a Condition has, but the class/method/arg info is now optional (unlike Conditions, where those three defined the Condition itself). This will allow Validators to have that info if available, or not. This means, unlike Conditions, Validators can be created directly, without a caller method from a type, allowing the user, if desired, to skip extending a Value and creating a method to return the Validator, letting them instead create a Validator directly.
 - Note that even though Validators have that functionality, it is not yet used directly (only indirectly in part through Conditions) and it is hidden. To actually take advantage of it, Conditions should be removed, since now all necessary info (and more) is inside Validator, and with a more flexible structure. However, to remove Conditions, major changes to Spec are needed. Since that is part of #510 and #511, Conditions will be kept until such moment, and the functionalities provided in this PR are just internal refactoring to allow that to be done later.


sei-aderr (commented):
Do the tests for validators not have to be adjusted based on these changes because these are updates that won't be utilized by other parts of the code yet?

mlte/validation/validator.py
```diff
@@ -49,36 +57,83 @@ def __init__(
...
+        caller_function: Optional[FrameType] = None,
+    ) -> Validator:
+        """
+        Creates a Validator using the provided test, extracting context info from the method that called us.
+
+        :param bool_exp: A boolean expression that can be used to test the actual condition we want to validate.
+        :param success: A string indicating the message to record in case of success (bool_exp evaluating to True).
+        :param failure: A string indicating the message to record in case of failure (bool_exp evaluating to False).
+        :param info: A string indicating the message to record in case no bool expression is passed (no condition, just recording information).
+        :param caller_function: A FarmeType with data about method that originally called this function. SHOULD BE REMOVED WHEN CONDITIONS ARE.
```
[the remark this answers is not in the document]
sei-secheverria: Fixed, thanks!

sei-secheverria:
> Do the tests for validators not have to be adjusted based on these changes because these are updates that won't be utilized by other parts of the code yet?

Very good question. On one hand, yes, they are not used by the code yet, or at least they are not exposed or useful yet, so we could postpone that. But on the other hand, I just forgot, and it makes much more sense to update/add unit tests when each module is created/updated. I'll add them and send an updated PR. Thanks!

mlte/validation/validator.py
```diff
@@ -49,36 +57,83 @@ def __init__(
...
+    @staticmethod
+    def build_validator(
+        bool_exp: Optional[Callable[[Any], bool]] = None,
+        success: Optional[str] = None,
+        failure: Optional[str] = None,
+        info: Optional[str] = None,
+        caller_function: Optional[FrameType] = None,
+    ) -> Validator:
+        """
+        Creates a Validator using the provided test, extracting context info from the method that called us.
```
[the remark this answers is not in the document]
sei-secheverria: So it would work exactly as it is working right now for Conditions. I didn't really change or add this functionality of extracting the context (calling class/method/args); Conditions were already doing that. I only 1) moved that functionality to its own class as a generic util, 2) added a way for the Validator to do the same that the Condition is doing by calling that now generic util functionality, but storing it in optional variables (whereas in Condition that information is kind of critical, as it defines the condition), even though for now this produces duplication, and 3) made it so the Condition.build_condition() method calls a new Validator.build_validator() method, both of which call that external generic functionality to gather the context info. 

That "caller_function" argument is only added temporarily, so that build_condition can pass that caller function that called it to build_validator. Once we remove Conditions, this arg would be removed as well. (If we didn't have this arg now, since build_validator is always called by build_condition, the context info of the caller function, from build_validator's perspective, would ALWAYS be "class: Condition, method: build_condition", since that is the only place where it is called, while what we want to store is WHO started the library call, which for now goes through build_condition, but would later be called directly. Btw, lines 90-94 are the ones that are "extracting context info from the method that called us", though that is not really used yet, as line 89, then 94, are the ones actually used since build_validator is always called from build_condition. Btw, you can look at the JSON file for a demo store to see how this looks like.

The idea is later to delete Conditions, and just call Validator.build_validator() directly. Why is this different than what we have with Condition? Because for Conditions, in practice the ONLY way to properly create them is to call build_condition(), which fills almost all of its fields with the context info. For Validators, it will be possible to either 1) call build_validator(), which is basically the same as Condition does with build_condition(), to make it transparent and allow storing that context info when needed, but it will also be possible to 2) call the Validator() constructor, and NOT pass any context info. This would allow someone defining a TestSpec/TestCase to create a new Validator on the fly, without having to create a ValueType class just to create a method for it that creates the Condition+Validator. In this case, the Validator would not have the context info, since it is a, if you will, "lambda-like" Validator that is just created on the fly. 

This change is not really done for library Value types, since in those cases it is easier to just call build_validator. This would be shown probably in the docs and in the demo, where it would show how Validators can be created on the fly if needed, or through whatever means you want. This depends on how the TestSpec/TestCase will be structured, so there may be further changes to the Validator when actually doing that, but hopefully not many.

sei-secheverria:
> > Do the tests for validators not have to be adjusted based on these changes because these are updates that won't be utilized by other parts of the code yet?
> 
> Very good question. On one hand, yes, they are not used by the code yet, or at least they are not exposed or useful yet, so we could postpone that. But on the other hand, I just forgot, and it makes much more sense to update/add unit tests when each module is created/updated. I'll add them and send an updated PR. Thanks!

Should be fixed

test/_private/test_reflection.py
```diff
@@ -0,0 +1,95 @@
...
+
+def test_lambda_and_comment():
+    lambda_with_lambda_in_comment = (
+        lambda w: w + 5
+    )  # this is a lambda  # noqa: E731
+    lambda_str = get_lambda_code(lambda_with_lambda_in_comment)
+    assert lambda_str == "lambda w: (w + 5)"
+
+
+def test_complex_source_line():
```
sei-aderr, line 70: What is the functionality that is causing the necessity for this test and `test_lambda_with_comment`? Why do other values like the comment, or the other param `new_str` end up in the lambda when you just send the lambda variable to `get_lambda_code`?

sei-secheverria: The context is that we want to store, for visual auditing purposes only, the code of the lambda used as the bool_expression for a Validator. That string is never really used by the system, but it could come in handy for someone reviewing how something was validated. The problem is that there is no easy way to get the code of a lambda. All methods are really bad, some worse than others. From what I can see, it could be possible to design a very robust method, but it could take quite some time, and the feature here is not so critical as to merit that amount of time.

That being said, all decent methods to get the lambda code require you to use inspect.getsourcelines. This module gets the lines of code where a given object was created. But the method is reeeeeally crappy, and it gives you a full line of code, instead of just the part of that line where the object was defined. Lambdas tend to be defined all over the place, so when you get that source line, you get the lambda hidden between who knows what else. Could be just the lambda, could be the lambda assigned to something, could be the lambda being passed dynamically as an argument to another function alongside other arguments (as is usually our case), sometimes without having the function name in that same line, sometimes with it, sometimes even with comments at the end of the line. So this, the best method to get started, is pretty awful.

Then you have to parse that potentially awful partial line of code. Since it may not be a full statement, it is very, very hard to figure out a proper regex that would work. For the same reason, it may be impossible to use a code parser to parse the lambda code out of it, as, the line being potentially part of a statement, it may not be parseable on its own. Of course, to fix that, you may get the previous lines, but you may need the next lines as well, and who knows how many more. Since the inspect module does not give you the full statement of where an object is defined, a robust way to get that lambda could take a long time.

So what I ended up doing is trying to parse the lambda out of the line of code. If we are lucky, and the line of code is parseable Python, we get the lambda nicely. If we are unlucky, and it is part of a longer statement in multiple lines, it won't be parseable. In that case, we simply find the "lambda" keyword and get whatever is after that, which may or may not contain things further down the line that are not part of the lambda. This can be a confusing result, since it will include the lambda plus some trash code, but it is the best compromise I could find.

The test for lambda with a comment, then, is to test when the line is parseable, that it was parsed properly and the comment was ignored. It may be kinda covered in some other tests, but I left it there since at some point, it was really needed to the way an older implementation I tested worked.

The last test is a case where it can't parse the line and ends up getting everything after the lambda keyword, which includes trash code line new_str. It should not be there, but it is how it is designed to work for now, and I don't think we can improve on that without a lot of effort.

sei-aderr: Got it. I didn't realize lambdas could be that much of a pain to work with. That all makes sense then and the PR looks good

sei-aderr (approved):

PR #591, opened 2025-01-30
First chunk of MLTE-505

sei-aderr:
Addresses several issues that have been merged into this branch in previous PRs. Now merging back into master


mlte/store/custom_list/underlying/fs.py
```diff
@@ -0,0 +1,136 @@
...
+        return entry
+
+    def _read_entry(self, entry_name: str) -> CustomListEntryModel:
+        """Reads a custom list entry."""
+        self.storage.ensure_resource_exists(entry_name)
+        return CustomListEntryModel(**self.storage.read_resource(entry_name))
+
+    def _write_entry(self, entry: CustomListEntryModel) -> CustomListEntryModel:
+        """Writes a custom list entry to storage."""
+        self.storage.write_resource(entry.name, entry.model_dump())
```
sei-secheverria, line 124: This is not a regression, but just a result of copying from a previous store version. But it should be changed to to_json

mlte/validation/validated_spec.py
```diff
@@ -119,11 +119,11 @@ def from_model(cls, model: ArtifactModel) -> ValidatedSpec:
 
     def print_results(self, type: str = "all"):
         """Prints the validated results per property, can be filtered by result type."""
-        if type not in ["all", "Success", "Failure", "Info"]:
+        if type not in ["all", "Success", "Failure", "Ignore"]:
```
sei-secheverria, line 122: This is another regression. "Ignore" no longer exists, it was changed a while ago to "Info".

test/custom_list/test_custom_list.py
sei-secheverria, line 1: This tests are ok, but I would add a test to convert from each model to the json, and back. There are equivalent tests (sometimes called "round trip") for other artifacts and entities you can copy from.

sei-secheverria (changes requested):
At least right now it does not look as messy as before. I found two regressions to fix, and a couple other minor things.

test/custom_list/test_custom_list.py
sei-aderr: Isn't that what these tests are doing? They create the objects from the model, then convert them to json, and then back to the model from json and ensure they are the same.

Or is there another flow of converting them that you are thinking. Looking at for example `test/model/test_shared.py` I could also add that initial json `deepequal` check and create an empty model of each but otherwise it looks to be doing the same thing

sei-secheverria: There are three different things: objects (which in this case is a vague term), models (in this case Pydantic models) and JSON. One test is to convert from the object to the model and back, another to convert from the model to json and back. However, you are right that, for the CustomList and entry, there are only models, so no additional tests are required. Things like Artifacts need both. So you can ignore this comment, I did not realize it does not apply in your case.

sei-secheverria (approved):

PR #596, opened 2025-02-04
MLTE-565 | Local Memory Custom List Store

sei-aderr:


mlte/store/custom_list/underlying/memory.py
```diff
@@ -0,0 +1,171 @@
...
+        self.storage = storage
+        """A reference to underlying storage."""
+
+    def create(
+        self,
+        entry: CustomListEntryModel,
+        list_name: CustomListName | None = None,
+    ) -> CustomListEntryModel:
+        self._check_custom_list_exists(list_name)
+        if list_name is not None:
```
sei-aderr, line 103: Open to feedback or suggestions on improving this check. I initially had it contained within `self._check_custom_list_exists` and had it raise an error if `list_name` was none, but the type checker does not accept that and complains that `list_name` could be none when using it as a key for the dicts.

mlte/store/custom_list/underlying/memory.py
```diff
@@ -0,0 +1,171 @@
...
+
+        raise ValueError("CustomListName cannot be None")
+
+    def read(
+        self, entry_name: str, list_name: CustomListName | None = None
+    ) -> CustomListEntryModel:
+        self._check_custom_list_exists(list_name)
+        if list_name is not None:
+            if entry_name not in self.storage.custom_lists[list_name]:
+                raise errors.ErrorNotFound(f"Custom List Entry {entry_name}")
```
sei-secheverria, line 135: I see these two lines repeated a lot. They are only slightly more complex than the equivalent in other memory store implementations, but maybe it would be better to move them to another private function, similar to check_custom_list_exists, to avoid having the check and exception message duplicated so many times.

test/store/custom_list/custom_list_store_creators.py
```diff
@@ -8,10 +8,29 @@
...
 from pathlib import Path
+from typing import Optional
 
 from mlte.store.base import StoreType, StoreURI
 from mlte.store.custom_list.factory import create_custom_list_store
 from mlte.store.custom_list.underlying.fs import FileSystemCustomListStore
+from mlte.store.custom_list.underlying.memory import InMemoryCustomListStore
+
+CACHED_DEFAULT_MEMORY_STORE: Optional[InMemoryCustomListStore] = None
+"""Global, initial, in memory store, cached for faster testing."""
```
sei-secheverria, line 19: This is not bad, though I think it may not be strictly needed in this case. The main reason I had to add this to the User store was because whenever you create a user, its password has to be hashed, and whenever we created a store, it came with default users. That hashing is expensive, and running that for every single test was consuming too much time and making testing too slow. For this store, creating a new store has no really expensive functionality, so we don't really gain much by this cache. I am not sure if I did it or not for catalogs, they are a different beast in terms of structure. In any case, this doesn't hurt. Up to you if you think it is worth it.

test/store/custom_list/fixture.py
```diff
@@ -29,35 +32,33 @@ def custom_list_stores() -> Generator[str, None, None]:
     """
     # for store_fixture_name in StoreType:
     #     yield store_fixture_name.value
-    yield StoreType.LOCAL_FILESYSTEM.value
+    for store_fixture_name in [
+        StoreType.LOCAL_FILESYSTEM.value,
+        StoreType.LOCAL_MEMORY.value,
+    ]:
+        yield store_fixture_name
```
sei-secheverria, line 40: Please add a TODO to explain what is going on and when it has to be changed (basically, when the new store types are added).

sei-secheverria (changes requested):
Ok for the most part, minor comments

mlte/store/custom_list/factory.py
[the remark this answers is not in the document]
sei-secheverria: Btw, just checked and, even before this PR, the fact that this is not handled in the customlist store does currently break the system if RDB is used for the artifact store. So it is even more important to address this, as the current master version breaks if the RDB is used.

mlte/store/custom_list/underlying/memory.py
sei-aderr: Addressed by going with the `check_valid_custom_list` option

mlte/store/custom_list/factory.py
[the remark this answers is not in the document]
sei-aderr: Yeah, would like to discuss as you would better understand the implications of each option.

If I set it up to return memory/filesystem if the param matches and then FS in any other case would that work fine? I don't remember how the catalog store handles creating a path for the FS store in the default case. 

Also, why does it break in master if it always returns a FS store? Does it just need a bit more handling to make things work

mlte/store/custom_list/underlying/memory.py
sei-aderr: Addressed

test/store/custom_list/custom_list_store_creators.py
sei-aderr: It is there for the catalogs, so I will just leave it to keep them consistent.

test/store/custom_list/fixture.py
sei-aderr: Added

test/store/custom_list/custom_list_store_creators.py
sei-secheverria: Fair enough

mlte/store/custom_list/factory.py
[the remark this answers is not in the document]
sei-secheverria: It breaks in master due to the URL. When it creates FileSystemStorage, it parses the URL to get the folder name, and it has a check to see, again, if it is a valid FS URL. Since it is not, when it is a DB, even if the FS store is properly created, it can't be fully initialized due to the incompatible URL being passed to it.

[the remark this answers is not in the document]
sei-secheverria: The sample catalog does a very weird thing, which is done when creating it specifically, so maybe not the best option. Maybe the simplest way would be to always override the URL with a hardcoded URL or something like that. We can discuss that in person if you want, it may be hard to write things down here.

sei-secheverria (approved):
Looks ok to me

Issue #599, opened 2025-02-21
Bug Writing Values to Report Results


sei-secheverria:
When writing a Value to the message of a Result, which are printed when passed as args to a validation to gain an audit of args used when validating, the resulting Result string has a full JSON of the Value, instead of the actual internal value. This is probably due to the json function used to generate the Result message, which, in case of Values, will generate the full JSON instead of getting the internal value (with str()). Modify so that args are first turned to strings, then into json arrays/dicts when writing them into the message of a Result.

sei-secheverria:
Addresd by PR #600

PR #600, opened 2025-02-21
Fixes Bug Printing Values to Result Messages in Validator

sei-secheverria:
Validator: fixed bug where messages set to generated Results when valdating were not properly serializing Values, turning them into JSON instead of getting their internal values. Addresses #599. Now args are turned to strings before the arrays are dumped to JSON, letting a Value properly be converted to its string represenatation.

### DIFF
--- a/demo/simple/confusion_matrix.py
+++ b/demo/simple/confusion_matrix.py
@@ -43,7 +43,7 @@ class ConfusionMatrix(ValueBase):
         return self._equal(other)
 
     def __str__(self) -> str:
-        return f"{self.matrix}"
+        return f"{self.matrix}".replace("\n", ", ")
 
     @property
     def misclassifications(self) -> int:
--- a/mlte/validation/validator.py
+++ b/mlte/validation/validator.py
@@ -125,20 +125,37 @@ class Validator:
                     "Configured bool expression does not return a bool."
                 )
 
-        # Stringify arguments so that result's message can include generic information about arguments used when validating.
-        values = f"- values: {json.dumps(args) if len(args) > 0 else ''}{', ' if len(args) > 0 and len(kwargs) > 0 else ''}{json.dumps(kwargs) if len(kwargs) > 0 else ''}"
-
         # Create the result to be returned.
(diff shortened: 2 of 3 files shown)


sei-aderr:
The validation of this working would be the report correct displaying results in the edit and also the view page? Both views look to be working

sei-aderr:
Actually besides this one which is showing some new lines that don't translate properly into the frontend
image

sei-secheverria:
> The validation of this working would be the report correct displaying results in the edit and also the view page? Both views look to be working

Right, the idea is that it shows up on those two cases.

sei-secheverria:
> Actually besides this one which is showing some new lines that don't translate properly into the frontend 

That is an interesting edge case I had not taken into account. Not sure if this has always been this way, btw. Now, I may argue that the current display is not that bad, as I don't think newlines should really be displayed as newlines in the report, but maybe they should show up as commas. I'll take a look to see if it is something easy to handle.

sei-secheverria:
> > Actually besides this one which is showing some new lines that don't translate properly into the frontend
> 
> That is an interesting edge case I had not taken into account. Not sure if this has always been this way, btw. Now, I may argue that the current display is not that bad, as I don't think newlines should really be displayed as newlines in the report, but maybe they should show up as commas. I'll take a look to see if it is something easy to handle.

Ok, handled it by changing the way Confusion Matrix's str() method works. It is a bit dangerous to handle it at the MLTE level, so it makes more sense for the specific value to handle it.

sei-aderr:
What is the output you are seeing for the report in the frontend? I am not seeing any change with that new commit

sei-secheverria:
> What is the output you are seeing for the report in the frontend? I am not seeing any change with that new commit

This is what I see now in the validated spec:

"Misclass count is less than threshold 2 - values: [\"[[10  0  0],  [ 0 10  0],  [ 0  0 10]]\"]" , which clearly does not have the ns. Note that you have to reload the confusion matrix (reload the Jupyer kernel), and re-run the evidence notebook for this to be properly executed, not just re-generate the report.

sei-aderr (approved):

PR #602, opened 2025-02-27
Py312 update

rbedman:
Update of demo notebooks and supporting cached model predictions to support python 3.12 and updated tensorflow model. Includes cleanup of notebook outputs for demos.


GALewis (approved):
The demo works great with Python 3.11.11. Thank you so much! The only comment is what I shared at the meeting today, which is not a major thing as long as I remember to do it. The first cell in the notebooks is an import. For some reason, if I use Run All, it does not execute this cell and therefore other cells fail. Once I execute that cell separately, I can use Run All and it works fine.

PR #603, opened 2025-02-27
MLTE-588

sei-aderr:


mlte/custom_list/custom_list_names.py
```diff
@@ -14,3 +16,32 @@ class CustomListName(StrEnum):
 
     QA_CATEGORIES = "qa_categories"
     QUALITY_ATTRIBUTES = "quality_attributes"
+
+
+class CustomListNameDict(dict):
+    def __setitem__(
+        self: "CustomListNameDict",
+        key: CustomListName,
+        value: Optional[CustomListName],
```
sei-aderr, line 25: I assume we want this value to be possible to be None, so this would be the correct way to handle that right?

Or I suppose we could just leave it without an entry too. Either way when you want to check the parent you have to either ensure there is an entry, or ensure it is not none

sei-aderr:
Have that one question, then just looking to figure out what mypy is expecting for the failure it is throwing with how I am overriding the dict class

mlte/custom_list/custom_list_names.py
sei-secheverria: Good question. I think it would be simpler to just check if there is no entry in the mapping, and force this not to be None. The benefit of having an explicit entry set to None is that you are explicitly stating that that list has no parent, but on the other hand, if one has to remember checking both that there is an entry and that it is not None, it makes it a bit easier to make a mistake. There is not definitive answer, but I tend to err towards simplicity, so I would change it to make value not None, and to remove the explicit QACategories=None from the mapping.

sei-secheverria (changes requested):
Everything looks good, suggestion in answer below.

sei-secheverria (approved):
Looks good

Issue #524, opened 2024-11-15
Add ability to run test cases within a TestSpec


sei-aderr:
Ideally a TestSpec would be able to be ran, and re-ran at any time. This will likely have many challenges with making the code in the TestCases available, but see what can be done

sei-secheverria:
As part of this ticket, also see if demo can be updated to showcase new functionalities: creating Validators on the fly, running these multiple tests together, etc., as far as possible without huge changes.

sei-secheverria:
Addressed in #608

Issue #512, opened 2024-11-15
Restructure ValidatedSpec into TestResults


sebastian-echeverria:
Restructure the current ValidatedSpec into TestResults. More specifically:

 - Rename to TestResults
 - Ensure that it does not contain duplicated info from TestSpec, but only the results, pointing to a specific TestSpec
 - Update backend stores to work with new structure

sei-secheverria:
Addressed in #608

Issue #509, opened 2024-11-15
Epic 8: New Structure for Test Cases


sebastian-echeverria:
Restructure Spec to be more clearly focused on test cases, more linear to set up and potentially easier to re-run.

sei-secheverria:
Completed.

PR #608, opened 2025-03-18
Feature/test cases

sei-secheverria:
Addresses most of epic #509 to refactor structure of Spec into TestSuite with TestCases and Validators. More specifically:

 - #510: Creates TestCase sub-artifact that structures information about tests, including QA Scenario it is addressing, as well as optionally the Validator and Measurements to be used for it.
 - #511: Restructure Spec into TestSuite, which removes Properties/QACategory objects from it, replacing it with a list of TestCases instead. Both TestSuite and TestCase were moved to a "mlte.tests" package.
 - #512: Renames ValidatedSpect into TestResults, though structure is kept vey similar. Frontend use of Report was updated to use new structure of TestSuite. Also added helper functions that load default TestSuite and all Evidence and validate them with one call.
 - #524: Added ability for TestSuite to run measurements for all TestCases that have it configured. If any of those are ExternalMeasurements, this will only work if the external function is readily available when running the tests method.
 - Moved String extended Evidence type from Demo to a new, top-level Evidence type.
 - More refactoring: Value was renamed Evidence, EvidenceMetadata was improved and moved into the renamed "mlte.evidence" package (former "mlte.value" package), and is now added more transparently to Evidences generated by Measurements.
 - Improved Validators to add support to store expected output type, and validate it when executing one
 - Demo: added tool to easily cleanup temporary outputs from notebooks, and simplified demo test scripts
 - Demo: removed most Evidence (formerly Value) types used in demo only to hold a method to create a Validator, and moved those methods to a validator module.
 - Removed QACategory (formerly Property) that are no longer used in TestSuite
 - Removed Conditions (were replaced by Validators)
 - Improved support to get fully qualified class/function names for Validators in different circumstances
 - Unified some duplicated Artifact methods into base class ones
 - Improved Measurements to fix minor bugs, reduce duplication, and simplify usage, as well as making most of it serializable
 - Plus several minor bug fixes and addition of missing unit tests


mlte/_private/function_info.py
```diff
@@ -54,24 +59,76 @@ def get_function_info(
...
+        function_name: Optional[str] = caller_function.f_code.co_name
+        if function_name == "<module>":
+            function_name = None
 
         # Get args. Args will include all caller arguments except for caller class or object.
         arg_keys, _, _, arg_values = inspect.getargvalues(caller_function)
         filtered_args = []
         for arg_key in arg_keys:
-            print(f"{arg_key}={arg_values[arg_key]}")
+            # print(f"{arg_key}={arg_values[arg_key]}")
```
sei-aderr, line 71: Is this intended to be left in?

mlte/_private/function_info.py
```diff
@@ -54,24 +59,76 @@ def get_function_info(
...
+            return ref
+    return None
+
+
+def get_class_from_func(func: Optional[FunctionType]) -> Optional[type]:
+    if not func or "." not in func.__qualname__:
+        return None
+
+    cls_qual_name, func_name = func.__qualname__.rsplit(".", 1)
+    assert func_name == func.__name__
```
sei-aderr, line 110: Is there a reason there is an assert here and not some other type of error handling?

sei-aderr:
Could you rebase/merge master into `releases/v2.0-staging`? Seeing a lot of changes in this diff relating to the custom list changes that merged into master

sei-aderr:
Also noticed a couple frontend display issues, made new issue #616 to track

mlte/tests/model.py
```diff
@@ -0,0 +1,36 @@
...
+
+class TestCaseModel(BaseModel):
+    identifier: str
+    """A name/id for the test case."""
+
+    goal: str
+    """A description of the goal of this test case."""
+
+    qas_list: list[str] = []
+    """A list of ids of Qualit Attribute Scenarios that this case is addressing."""
```
sei-aderr, line 21: Typo: Quality Attribute Scenarios

test/_private/test_meta.py
```diff
@@ -0,0 +1,83 @@
...
+# Tests
+# -------------------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize(
+    "test_object, callables, expected",
+    [
+        (ClassForTesting, ["method_for_test"], True),
+        (ClassForTesting, ["not_method_for_test"], False),
+        (ClassForTesting, ["method_for_test", "a_class_method"], True),
```
sei-aderr, line 46: Should this also test the static method?

sei-aderr:
Generally looks good, don't have time to nitpick everything though due to the incredibly large diff. Had a couple comments, fixed a couple myself. Should be good to merge with those addressed

sei-aderr (approved):

mlte/_private/function_info.py
sei-secheverria: Debatable. Not really, it helped in debugging and may help again, but it is not really needed.

mlte/_private/function_info.py
sei-secheverria: Good question. Most likely that I got that code from somwhere and didn't change it :)

sei-secheverria:
> Question: Are we planning to load in the `Quality Attribute Category` field once the frontend/backend linking are done? I assume that is the case and why you left those `<td>` empty and left the column

Yes, that is the idea.

mlte/_private/function_info.py
sei-secheverria: Addressed

mlte/_private/function_info.py
sei-secheverria: Addressed

test/_private/test_meta.py
sei-secheverria: Wasn't mandatory as the static method is used for another test, but it is not a bad idea to do it here as well. Addressed.

mlte/negotiation/artifact.py
```diff
@@ -64,8 +65,14 @@ def to_model(self) -> ArtifactModel:
         )
 
     @classmethod
-    def from_model(cls, model: ArtifactModel) -> NegotiationCard:
+    def from_model(cls, model: BaseModel) -> NegotiationCard:
         """Convert a negotiation card model to its corresponding artifact."""
+        assert isinstance(
```
sei-aderr, line 70: This is merged now, but you changed one assert I mentioned while reviewing but there are still these two and then  one in the `local_object_size.py` should these be adjusted too or were these ones intentional?

sei-secheverria: These asserts are different. It wasn't intentional or not, it was just there from before. Kyle liked to use asserts for some cases, for some reason, instead of explicitly raising exceptions. I just haven't taken the time to change those, and for consistency I've kept them like that.

Issue #607, opened 2025-03-17
Update docs to reflect changes to TestSuite, TestCases, TestResults


sei-secheverria:
From changes in #510, #511 and #512, the docs have to be updated for consistency. Check changes in branch "feature/test-cases" for the current new version of the code. The main changes are:

 - Spec is now called TestSuite
 - TestSuites are not organized by Property, but they are just a list of TestCases, where each case has a link to the QAS that it is addressing (note that changes on the NegotiationCard so that QASs can have these ids are being handled in a separate Epic, #505, which has a similar ticket for updating docs, #518; however, this epic is not ready yet).
 - ValidatedSpec is now TestResults, but other than that it is similar to how it was

sei-secheverria:
Addressed in #625

Issue #574, opened 2025-01-13
Improve Coherence of Documentation


sei-secheverria:
There are some issues with these two related docs:

 - Readme is way too short. It should include at least general information about using MLTE and setting up the dev env, from the corresponding docs.
 - "Setting Up" is not a good name for that doc. There is barely any setup, most of it is about how to use MLTE, more of a  "Using MLTE" (note overlap with existing Using MLTE doc). Rename for clarity.
 - See how to simplify the current Using MLTE doc, and potentially merge it with MLTE Process
 - index of docs also has overlap with these two. See how to optimize this.
 - Review the "Refernces" docs for use and consistency as well

sei-secheverria:
Will be done along #607

sei-secheverria:
Addressed in #625

PR #625, opened 2025-04-07
Update Documentation (for Coherence and New MLTE Functionality)

krmaffey:
Updates documentation for coherence and to describe new MLTE functionality - addresses issues #607 and #574. Significant changes include:
- The MLTE framework is on the documentation landing page
- Using MLTE spells out how to use the Python package with mention of how that fits into the framework
- The GitHub README includes brief instructions on how to install and use MLTE, with links to corresponding documentation

When reviewing, there are two outstanding questions for the team:
- Do we still have any use for the resources page within the documentation? Verbiage was updated to reflect QAs but it's unclear if anyone will find the page useful.
- Should we maintain the MLTE measurements page? Or assume that users can reference the API documentation that's auto-generated?

Looking forward to your edits and comments!


README.md
sei-secheverria, line 1: This looks very good. The only thing I am wondering if we should link here is something about the demo, which is the easiest way to understand how mlte works. Not sure if a summary would be appropriate, since it may be to long to put here; maybe a link to the demo section in the rest of the docs to point people to the right location?

docs/docs/using_mlte.md
```diff
@@ -1,285 +1,225 @@
...
-- Teams design a specification by selecting and prioritizing the model requirements that are important to their project from the list of `MLTE` [*properties*](properties.md). 
-- To validate that a requirement for a property is met by an ML model and system, `MLTE` uses *measurements*, which correspond to properties. Selecting measurements that correspond to properties is part of `Spec` definition.
+1. Define a `TestSuite`.
+3. Collect `Evidence`.
+4. Validate results.
+5. Examine findings.
+
+### 1. Define a `TestSuite`
+
+A `TestSuite` is an artifact made up of a collection (or suite) of tests that the completed model must pass in order to be acceptable for use in the system into which it will be integrated. The `MLTE` `TestSuite` is  made up of smaller units, which we call a `TestCase`. Each `TestCase` includes a name or title, a goal, a link to a corresponding quality attribute in the Negotiation Card, a threshold, and finally a measurement. Here is an example of a `TestCase`:
```
sei-secheverria, line 159: I would explicitly indicate here that the measurement is optional, since it can be ignored, and then created and executed later.

docs/docs/using_mlte.md
```diff
@@ -1,285 +1,225 @@
...
+        validator=Real.greater_or_equal_to(0.98),
+        measurement=ExternalMeasurement(
+            output_evidence_type=Real, function=accuracy_score
+        )
 )
-spec.save(parents=True, force=True)
 ```
 
-### Collecting Evidence
+To see a simple example of a `TestSuite`, refer to the `2_test_suite.ipynb` notebook in the <a href="https://github.com/mlte-team/mlte/tree/master/demo/simple" target="_blank">simple demo folder</a>. To see an example of a more extensive `TestSuite` including quality attribute scenarios, see the requirements notebook in the <a href="https://github.com/mlte-team/mlte/tree/master/demo" target="_blank">scenario demo folder</a>. All demos and corresponding instructions can be found in the <a href="https://github.com/mlte-team/mlte/tree/master/demo" target="_blank">demo folder</a>.
```
sei-secheverria, line 173: The "scenarios" demo link is missing the final "scenarios" folder in the URL.

docs/docs/using_mlte.md
```diff
@@ -1,285 +1,225 @@
...
 
-After building the `Spec`, teams must collect evidence to attest to whether or not the model realizes the desired properties. Here we show one way to collect evidence. Note that this example does not include data and model training code, but those can be found in the full `MLTE` <a href="https://github.com/mlte-team/mlte/tree/master/demo" target="_blank">demo notebooks</a>.
+### 2. Collect `Evidence`
+
+After building the `TestSuite`, `MLTE` requires developers to collect evidence to attest to whether or not the model realizes its stated goals. `Evidence` is generated by the `Measurement`s which are specified in each `TestCase` as described [above](#1-define-a-testsuite). Each invididual piece of data produced by running a `TestCase` is an `Evidence`. Once `Evidence` is produced, we can persist each one to an artifact store to maintain them across sessions.
+
+Here is an example of collecting `Evidence` for a `TestCase` on accuracy:
```
sei-secheverria, line 179: Following from the comment above, this is currently a bit confusing. It indicates that evidence is obtained from the Measrurement specificed in a TestCase, and the the example below does not have anything to do with the measurement specificed in a TestCase. This is because now you can either specify the measurement in the test, and then just pass all inputs and run run_measurements (as seen in simple demo), OR not specify one or more measurements in the testcase, and do what we are explaining here below. This should probably be clarified, the fact that there are two workflows for this, and this is one of them

sei-secheverria: Maybe we should explicitly explain both flows, with different examples

sei-secheverria (dismissed):
Did a quick review, for the most part I think it looks great. Main thing I'd do is to clarify the two worklfows for measuring, when doing it manualy or automatically, which are reflected in the two demos.

Regarding your questions:
 - The Resources page seems outdated. The default QA categories and QAs that we will include will be changed relatively soon, and I guess part of the goal of the TestCatalog is to replace a bit what was going on here. We should probably discuss it.
 - The measurements page, it is a good question. We do not have many built-in measurements, but they are not terribly easy to find in a list in the API. It may be useful to keep something like this. We should discuss it as well.

README.md
krmaffey: Great point - added a short summary paragraph at the top which includes links to the framework, using_mlte, and the demo folder

docs/docs/using_mlte.md
krmaffey: Added!

docs/docs/using_mlte.md
krmaffey: Excellent catch thank you, updated the scenario demo folder URL

docs/docs/using_mlte.md
krmaffey: You raise great points here - I don't think the section makes sense the way it is currently written, but I'm not totally sure what the best way is to modify it. Leaving this unresolved for now, will come back to it after some more thought.

krmaffey: Okay, thanks for your help the other day - I've added in some more details on the two workflows and linked to the demos. Let me know if you think it needs to be adjusted further!

sei-secheverria: Looks great now, thanks!

sei-secheverria:
We are at a point where it is much simpler to merge this into the staging branch, so I'll be merging this in. If we want, we can have another review at the doc updates separately and fix any issues later.

Mover remaining questions to #656 and #657

Issue #517, opened 2024-11-15
Frontend: Change Negotiation Card View to select from QA CustomList instead of having a text input


sei-aderr:
Make the QAS in the frontend use the list of QAs instead of the text input it has. More specifically:
 - The frontend should use the CustomList API to get the list of QAs
 - The View/Edit page for the NegotiationCard should now show a dropdown menu in the Quality Attribute section, instead of a text field, and show the options from the CustomList
 - It is very likely we will need to add 2 dropdowns: one for the QA category, and one for the QA, though we probably just want to store the QA with the negotiation card. This would mean ensuring in the frontend that we load and select the corresponding parent category when displaying the stored QA from a stored NegotiationCard.
 - The sample negotiation card in the demo should be loadable in the Frontend, but it is very likely that the QAs in the negotiation card do not match the ones currently in the default CustomList for QAs. Discuss as needed to align them.

sei-secheverria:
Addressed in #634

Issue #587, opened 2025-01-27
Change Negotiation Card QASs to use CustomList entries instead of strings for QAs


sei-secheverria:
Associate QAs stored in the QA CustomList to a QAS, instead of the string it has now. More specifically:
 - Change the NegotiationCard artifact to store CustomEntries instead of a string; likely same for the model.
 - Ensure that the Backend endopoints work properly with this change. #517 is the corresponding change for the Frontend.
 - If needed, change ArtifactStore as well (the RDBS implementation may need updating; though note that it will NOT link to the CustomLists table, as they are in separate stores).

sei-aderr:
No longer to be addressed, because the frontend is already able to handle the getting all of the info about custom list entries so it doesn't need it from the negotiation card directly.

We don't want to change it because the negotiation card because the we don't want to have data from the custom list store in the artifact store in a negotiation card because they are not able to be directly linked.

So because this is handled in the frontend, we need to add support for a  library session to be able to also see the list of custom list entries which will happen in a separate issue.

sei-aderr:
Continued in #650

PR #634, opened 2025-04-22
Feature/frontend-qas-link

sei-aderr:
Addresses most of #517 and has a fair bit of random linting done by the linter that didn't get done before.

Does not address making sure that the demo negotiation card can be loaded as the QA in the example are not currently part of the sample catalog and so will not be populated in the drop downs. Would make the most sense to fix that once the rest of the team finishes their set of QAC/QA, so can make a separate issue for that, just can't forget to do it before the 2.0 release. 

Although, we did set the milestone for test catalog to be in August, well after the 2.0 release. So I suppose I could populate the default custom lists with the values needed for the example to work for now, and then update it later. What do you think makes the most sense?

### DIFF
--- a/demo/sample_store/models/OxfordFlower/0.0.1/default.negotiation_card.json
+++ b/demo/sample_store/models/OxfordFlower/0.0.1/default.negotiation_card.json
@@ -201,7 +201,7 @@
             "system_requirements": [
                 {
                     "identifier": "qas_id_001",
-                    "quality": "Fairness: Model Impartial to Photo Location",
+                    "quality": "Model Impartial to Photo Location",
                     "stimulus": "The model receives a picture taken at the garden",
                     "source": "the flower identification application",
                     "environment": "normal operations",
@@ -210,7 +210,7 @@
                 },
                 {
                     "identifier": "qas_id_002",
-                    "quality": "Robustness: Model Robust to Noise (Image Blur)",
+                    "quality": "Model Robust to Noise (Image Blur)",
                     "stimulus": "The model receives a picture taken at the garden, and it is a bit blurry,",
                     "source": "the flower identification application",
                     "environment": "normal operations",
@@ -219,7 +219,7 @@
                 },
                 {
                     "identifier": "qas_id_003",
-                    "quality": "Robustness: Model Robust to Noise (Channel Loss)",
+                    "quality": "Model Robust to Noise (Channel Loss)",
                     "stimulus": "The model receives a picture taken at the garden",
                     "source": "the flower identification application running on a loaned device ",
                     "environment": "normal operations. These devices are known to sometimes lose a channel (i.e., RGB channel)",
@@ -228,7 +228,7 @@
                 },
                 {
                     "identifier": "qas_id_004",
-                    "quality": "Performance: Resource Consumption on Operational Platform",
+                    "quality": "Resource Consumption on Operational Platform",
                     "stimulus": "The model running on the loaned device receives pictures taken at the garden",
                     "source": "the flower identification application",
                     "environment": "normal operations. The loaned devices are small, inexpensive devices with limited CPU power, as well as limited memory and disk space (512 MB and 150 MB, respectively)",
@@ -237,7 +237,7 @@
                 },
                 {
                     "identifier": "qas_id_005",
-                    "quality": "Interpretability: Understanding Model Results",
+                    "quality": "Understanding Model Results",
                     "stimulus": "The model receives a picture taken at the garden",
                     "source": "the flower identification application",
                     "environment": "normal operations",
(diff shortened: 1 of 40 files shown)


sei-secheverria:
> Does not address making sure that the demo negotiation card can be loaded as the QA in the example are not currently part of the sample catalog and so will not be populated in the drop downs. Would make the most sense to fix that once the rest of the team finishes their set of QAC/QA, so can make a separate issue for that, just can't forget to do it before the 2.0 release.

> Although, we did set the milestone for test catalog to be in August, well after the 2.0 release. So I suppose I could populate the default custom lists with the values needed for the example to work for now, and then update it later. What do you think makes the most sense?

Yes, in theory I agree, but as you say, in practice v 2.0 and the update of the QA attributes are not lined up. We can't really leave the demo broken, so we will have to fix the demo negotiation card somehow in this PR. I would change the negotiation card to use whatever values we can for this to work. Probably the best option is not to change the demo negotiation card, but to add the QAs in the negotiation card to the list of pre-defined QAs that come with MLTE. We will change that later anyway, but at least now it will be consistent. You can try to figure out what category would apply to them, and also add those if needed, and I can review if I think they make sense.

sei-secheverria (changes requested):
Mostly looks ok, in terms of functionality. Some issues:
 - As already mentioned, demo negotiation card should work, and QAs used there should be in current, temporary default lists
 - Besides the above, in the current QA lists with have a ton of categories, but most of those categories do not have QAs in them. This is inconsistent and confusing. I would prefer to trim down the currently available categories, and only have some that do have QAs. This probably means removing most categories, and maybe adding some QAs to some of the categories that would not be deleted.
 - We are getting to a point where not having RDBS support for CustomLists, nor making them work as FS store if RDBS is used for other stores, can become a problem. The default docker deployment uses RDBS, and a lot of things are going to start kinda silently failing due to this. We should discuss this topic again.

sei-aderr:
Addressed the sample negotiation card and empty categories. All QAC/QA used in the demo scenarios and sample test catalog are there and loading in correctly. All QAC without children have been removed until the final list is created by research side.

Yes, that is a good point. It is something that will start popping up more and shouldn't be released with it broken in 2.0 but will make the end of April deadline significantly tougher.

sei-secheverria:
We may want to do the intermediate patch case, force customlists to always be a FS based store, though some URI-wrangling would be needed.

sei-secheverria:
I am missing something here... #517 depends on #587, which we have not implemented yet. I can't test if this is loading the proper QAS from a NegotiationCard without that backend change. Are you working on that one in separate? If so, that one should be reviewed/merged before this one.

sei-aderr:
> I am missing something here... #517 depends on #587, which we have not implemented yet. I can't test if this is loading the proper QAS from a NegotiationCard without that backend change. Are you working on that one in a separate branch? If so, that one should be reviewed/merged before this one.

So I had seen that change as coming after this one. Currently the frontend loads the QAC/QA with all of their data, it just presents only the names and then sends only the names. Because the frontend does not have the concept of the object of a CustomList entry in the same way the backend does, I was planning for the backend to accept the string value of the id of the entry and then convert that into the `CustomListEntry` object and then storing it. 

It would be possible for the frontend to store all of the data for the `CustomListEntry` and then pass it to the backend, but then the backend would still need to validate that the entry provided matches an entry it already has saved. Because if it doesn't we can run into issues where we are saving entries that aren't created with the proper CRUD and could not exist in the lists.

sei-secheverria:
Ok, sure, you could do that, and that is a valid approach. The issue that I see in the current branch is that when I load a NegotiationCard, the QA/QAC dropdown are not updated to select the QA that the NegotiationCard has, nor to find the parent QAC and select it as well. Is that an expected behavior/limitation of the current branch, to be fixed as you do the next part?

sei-aderr:
Could you try and pull again and check if it works? Something went wrong with it in the initial PR but I thought I had fixed it in some of the more recent commits. It was populating both parent QAC and QA for me for all both of the demo negotiation cards

sei-aderr:
> We may want to do the intermediate patch case, force customlists to always be a FS based store, though some URI-wrangling would be needed.

It seems that it will be necessary. Made task #643 to address

sei-secheverria:
I may be doing something wrong, but whenever I try running the run_enviornment.sh from demo/scenarios, which is rebuilding/using the latest version in the docker containers it is creating, I don't see the dropdowns being selected for the demo negotiation card that is copied there when I use the UI and go to it, the first one shows "Select", and the second one is empty. I seem to be rebuilding everything, so not sure why this wouldn't work for me only.

sei-aderr:
I am surprised there are even any options loaded in. Isn't the docker version using an RDBS store so the custom list store isn't created?

sei-secheverria:
Two things:
 - The docker deployment scripts can easily use the FS stores with some param changes. It's just that by default they use Postgres.
 - The scenarios demo script does not use Postgres but a FS store (setting some of those params), and that is what I've been testing, because unlike the default docker scripts, 1) it comes with a built in negotiation card to test (which is copied over each time the script starts, so it is always up to date), and 2) it uses FS by default, which is the target we are testing for the custom lists.

sei-aderr:
Got it. I see the same behavior you are seeing. Will look into it

Issue #633, opened 2025-04-22
Bug updating policies of models when logging in


sei-secheverria:
The backend has a functionality where it attemps to add missing permissions for models that are created through the library (since model permissions are only created when the UI is used to create the model). However, there seems to be a bug where a model created through the UI has proper permissions, but the backend think it doesn't, leading it to try to create the permissions when the user is logging in, and blocking log in as it crashes due to the permissions already being there.

sei-secheverria:
Addressed in #642

PR #642, opened 2025-04-24
Bug/log in model permissions

sei-secheverria:
Addresses #633 by adding explicit support for hyphens in resource ids, permissions were not working properly when a resource id contained a hyphen. Also added unit tests for the Policy class.


mlte/user/model.py
```diff
@@ -189,6 +189,10 @@ def get_group_names(groups: List[Group]) -> List[Group]:
 class Permission(BaseModel):
     """Permissions for manipulating resources."""
 
+    SEPARATOR: ClassVar[str] = "-"
```
sei-aderr, line 192: Is there a reason this one is defined as a ClassVar and the policy one is not?

mlte/store/user/policy.py
```diff
@@ -195,12 +199,20 @@ def save_to_store(self, user_store: UserStoreSession) -> None:
...
         for group in self.groups:
             for permission in group.permissions:
-                user_store.permission_mapper.delete(permission.to_str())
+                # Store permissions in dict for later removal, to avoid trying to re-remove already deleted ones.
+                permissions[permission.to_str()] = True
 
             user_store.group_mapper.delete(group.name)
 
+        # Now remove all permissions.
+        # TODO: note that this main leave other groups using these permissions dangling. Not trivial to check if
```
sei-aderr, line 211: This is a tough one. Be sure to make a task relating to this so that we don't forgot to look into it later. Feels like it should be a solved problem in the world of permission management but not immediately clear what that solution would be.

test/store/test_policy.py
```diff
@@ -0,0 +1,151 @@
...
+from test.store.artifact.fixture import (  # noqa
+    rdbs_store as artifact_rdbs_store,
+)
+from test.store.user.fixture import fs_store as user_fs_store  # noqa
+from test.store.user.fixture import memory_store as user_memory_store  # noqa
+from test.store.user.fixture import rdbs_store as user_rdbs_store  # noqa
+from test.store.user.fixture import user_stores
+
+
+@pytest.mark.parametrize(
```
sei-aderr, line 28: Didn't know you could stack them like this to get all of the combinations. A nice implementation

sei-aderr: Is the TestPolicy defined as a class so that you can pass this set of stacked `parametrize` to all of the static methods underneath without having to repeat the `parametrize` each time?

test/store/test_policy.py
```diff
@@ -0,0 +1,151 @@
...
+    @staticmethod
+    @pytest.mark.parametrize("store_fixture_name", user_stores())
+    def test_save_remove(
+        resource_type: ResourceType,
+        id: str,
+        read: bool,
+        edit: bool,
+        create: bool,
+        store_fixture_name: str,
+        request: pytest.FixtureRequest,
```
sei-aderr, line 76: What is this pytest object being used to get a UserStore?

mlte/user/model.py
sei-secheverria: Yes, when you are extending a Pydantic model, which we are here by deriving from BaseModel, every attribute becomes a model field that will be stored and has to be handled in constructor and serialization. To mark an attribute as a class attribute and not a part of the model to be serialized, this is the way Pydanting does it.

mlte/store/user/policy.py
sei-secheverria: Yep, I'll make a task for that. I have some ideas, it just would be a substantial amount of effort, and there are currently no clear use cases where this would actually happen (yet).

test/store/test_policy.py
sei-secheverria: Right. I looked into ways to avoid repeating the parametrized args and to reuse them for several, but not all, tests in a module, and this seemed to be the simplest option. There are other ways, but they either apply to all tests in the module, or have other downsides.

test/store/test_policy.py
sei-secheverria: We are doing that in several ways, it is one of pytest weird built in ways to do stuff. In this case, this allows us to pass store names as strings, and the magical request object lets us create objets that have been decorated as fixtures out of those strings. I basically copied that pattern from code Kyle had already added, there may be a more efficient  or clearer way to do it, but this is how we are doing it everywhere we want to test something with multiple store implementations.

sei-aderr: Is there a functional difference to this implementation to get the stores and how I set it up for the custom lists and the store I copied that from? Those ones just make the list of the store types and then have one function as a switch for each type. It looks very similar but could be a difference I'm missing

sei-secheverria: I had not realized it, but we do have two methods. The catalog and custom lists store are using the switch you mention, and the artifact and user ones are using the fixture call one. I am not entirely sure why we ended up with both versions, likely not intentional. It would be interesting to decide which one is better. It could be argued that this option (with the pytest.RequestFixture) is a bit harder to understand at first sight on each test, but it requires less code, as it doesn't require that switch (the switch is implicit in the fixture name and using request to call that fixture by its name), and is easier to scale. The other, switch option is simpler to follow when looking at a test, but it requires that switch, which means maintaing more code that may not be needed. We could discuss this in more detail.

sei-aderr: Yeah, can see the arguments for both. I don't have a super strong opinion either way. I am find to leave this one as-is and then just make a task to evaluate and make them consistent at a later date.

sei-aderr:
So in the end the issue was not related to how the permissions were created (library/frontend) but instead just an error with the naming of the artifacts that was causing issues and creation method wasn't an issue?

sei-secheverria:
Yes, the issue was hyphens in resource names. The groups and permissions are named automatically based on several things, including resource ids when they are groups/permissions for specific resource ids. The way I was building those names was based on hyphens, and when there were hyphens in the id, the split would not work properly, and some methods would fail to work as expected. Adding support to replace hyphens in resouce ids with another character has fixed the issue.

sei-aderr (approved):

Issue #626, opened 2025-04-14
Frontend: show QAS Ids in Negotiation Card View


sei-secheverria:
Once #516 is completed, modify the frontend's view of the NegotiationCard to also show QAS ids. This should be a read-only field, since they will be automatically generated. They will be empty when a new negotiation card is created, or when editing and a new QAS is added, but should show the "identifier" attribute of the NegotiationCard model when an existing NegotiationCard is loaded.

sei-secheverria:
Addressed in #644

PR #644, opened 2025-04-24
Add display for QAS ID in negotiation card

sei-aderr:
Addresses #626 

- Added a display for the identifiers in the frontend that works when they are pre-loaded like in the demo negotiation card. - Displays a message that ID is not yet defined when a new card is made

- Having issue with ID being populated when card is saved though. 
    - Without any changes to the format the frontend is using for the card, the identifier comes back as null. This format does not include the identifier in the object that is submitted to the backend
    - Changing the negotiation card to have an empty string by default and pass that to the backend results in the value for the ID remaining as an empty string after being saved.

It looks like from the changes in #628 that the backend is expecting the field to not be there, but don't see exactly what is going wrong in the generation to make it be null.


sei-secheverria:
Fixed bug that was not generating ids when card was saved by the backend.

Note that there is still a bug loading this data. In my tests, ids from loaded negotiation cards was not showing up, even though the frontend was receiving the ids. Now, when creating a new card, the frontend is receiving the new ids, but they are not showing up in that case either.

sei-aderr:
One note, when deleting a QAS and then adding a new one, the ID value will keep incrementing. So ID 1,2,3 when deleting 2 becomes 1,3,4. 

We probably want it to be this way so that if the user writes test cases for specific IDs, those IDs don't change right? Obviously if deleted it won't work anymore but this does prevent breaking every ID after the one that is deleted

sei-secheverria:
> One note, when deleting a QAS and then adding a new one, the ID value will keep incrementing. So ID 1,2,3 when deleting 2 becomes 1,3,4.
> 
> We probably want it to be this way so that if the user writes test cases for specific IDs, those IDs don't change right? Obviously if deleted it won't work anymore but this does prevent breaking every ID after the one that is deleted

That is a very good point, I hadn't considered deletion of QAS. Since we are managing these ids internally for the most part, I agree that we probably want this behavior. It may be the case that they delete QAS at the beginning, before any test cases are written, in which case there will be some discontinuity, but if it really bothers them they can recreate the negotiation card or fix it manually. But I agree that, if QASs are for any reason deleted later during the process, it is better to treat them as unique requirement ids, and treating a deleted QAS as a deleted requirement. If a requirement is deleted, they'll have to see what to do with the cases.

Note, however, that this is not currently perfect: if the last QAS is deleted, and a new one re-added, depending on the steps taken for this, it may end up with the same id as the deleted one. Not sure if we really have to handle that case, since that would require storing somewhere else the "highest" id ever used, regardless of the current ones.

sei-aderr:
Yeah, I don't think that saving the highest is a necessary implementation for now. I think we can go with how it is for now, make a task to consider it later and see if it comes up as people keep using it as an issue.

sei-secheverria:
Let me review your last change.

Issue #643, opened 2025-04-24
Implement work around for RDBS Custom Lists


sei-aderr:
Discussed in #634

Custom lists currently don't support RDBS/HTTP stores. The docker version uses the RDBS so it will start to have issues using the custom list items in the 2.0 release unless it is implemented, or a workaround is used to default use FS.

Plan is to make a workaround to use a currently implemented store to reach the 2.0 release goal of end of April and then the full store implementation can be finished later.

sei-secheverria:
Addressed in #647

PR #647, opened 2025-04-28
Feature/custom list rdbs workaround

sei-secheverria:
Addresses #643 by creating a default folder and setting a default file-system URI for CustomLists if DB or HTTP general store URIs are used.

### DIFF
--- a/mlte/backend/core/config.py
+++ b/mlte/backend/core/config.py
@@ -51,11 +51,11 @@ class Settings(BaseSettings):
             ) from None
         return v
 
-    STORE_URI: str = StoreURI.get_default_prefix(StoreType.LOCAL_MEMORY)
+    STORE_URI: str = StoreURI.create_uri_string(StoreType.LOCAL_MEMORY)
     """The store URI string; defaults to in-memory store."""
 
     CATALOG_URIS: Dict[str, str] = {
-        "local": StoreURI.get_default_prefix(StoreType.LOCAL_MEMORY)
+        "local": StoreURI.create_uri_string(StoreType.LOCAL_MEMORY)
     }
     """The dict of catalog URI strings; defaults to one in-memory store."""
 
--- a/mlte/backend/main.py
+++ b/mlte/backend/main.py
@@ -6,7 +6,7 @@ Entry point for MLTE artifact store server.
 
 import logging
 import sys
-from typing import Any, Dict, List
+from typing import Any
 
 import uvicorn
 from pydantic.networks import HttpUrl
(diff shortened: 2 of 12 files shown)


sei-aderr:
Looking into potential issue I noticed. When I ran the docker version in scenarios with `run_environment.sh`, `STORE_TYPE` set to rdbs it looks like there is a store created both at `/mnt/app/stores` with a shorter list of `qa_categories` and these are the ones that the UI sees. Then in `/mnt/store` there is another store structure with the full list of `qa_categories` and these are not all shown in the UI dropdown. 

Saw something similar when using `deployment/start.sh` but in that case there is the shorter list in `mnt/app/stores` and `/mnt/store` is empty

sei-secheverria:
> Noticed an issue with the store creations. When I ran the docker version in scenarios with `run_environment.sh`, `STORE_TYPE` set to rdbs it looks like there is a store created both at `/mnt/app/stores` with a shorter list of `qa_categories` and these are the ones that the UI sees. Then in `/mnt/store` there is another store structure with the full list of `qa_categories` and these are not all shown in the UI dropdown.
> 
> Saw something similar when using `deployment/start.sh` but in that case there is the shorter list in `mnt/app/stores` and `/mnt/store` is empty

I'll look into it. Btw, sure none of those were leftovers from previous runs? When I tested the run_environment, it seemed to load everything well in the NegotiationCard, but I didn't really check if the list was the same as expected.

Probably the issue has to do with the default folder I am using, and how it maps inside the container. For the second case, deployment/start, it uses a FS from the start, specifying a common folder, so it is understandable that the other one is empty (not sure why it is even there). Not sure I see why there is a shorter list, other than it being leftovers from previous runs.

sei-aderr:
The extra folder is there in the docker folder one is made by `docker/Dockerfile.backend`. It makes `/mnt/store` in that setup dockerfile.

It looks like the duplicated lists in `run_environment` may have been a leftover as I pruned the docker system and now don't see the second list. Only the one in `/mnt/app` but again it is the shorter list.

sei-secheverria:
The behavior I see is different:
 - When running run_environment, I see a /mnt/store folder, with all the expected custom lists (8 categories and 14 qa). This folder is mounted as a volume and points at demo/scenarios/store. I also see a /mnt/app/mnt/store folder that is empty, not sure why, seems to be a bug somewhere that it got created. /mnt/store is the folder created for the file system store, so other than the empty ghost folder, behavior is as expected.
 - When running the deployment/rebuild_and_start script, the /mnt/store is empty, which is expected as it is not being used, and I see a /mnt/app/stores folder with bot the sample catalog and the initialized custom lists, which is because that is the default folder name used in both cases when the main URI is not a FS. All 8 categories and 14 QAs show up there as well. So, other than the fact that calling it "stores" is a bit inconsistent, it works as expected.

I'll look into the ghost mnt/store subfolder, and see if it is more consistent to call the other one stores, but I do not see the behavior you say, with less attributes. I can only assume those are leftovers from previous versions.

sei-secheverria:
Ok, removed ghost /mnt/app/mnt/store, made clearer /mnt/app/default_store for forced FS stores, and refactored StoreURI a bit to remove duplication, adding built-in method for default FS store.

sei-aderr (approved):

Issue #627, opened 2025-04-14
Frontend: Show Quality Attribute and QAS Id in Report


sei-secheverria:
The Report currently has a column for the results called "Quality Attribute Category". We need to show the proper data here. More specifically:
 - Each TestCase will have one or more QAS ids associated to it. Using that id, the frontend will have to find the corresponding QASDescriptor in the NegotiationCard data section of the report, and get the "quality" field from it. That is what should go in the cell. 
 - If we also want to show the category, we have to get its parent, and mix the results.
 - We should also show the QAS id in the Report. We either mix those and show everything in the same column (something like "Quality Attribute Scenario" for title, and the "qas_001 - Robustness" for a value), or create different columns for each.
 -  Note that a test case may be associatged to more than one QAS. The previous would have to be done for all QAS ids, and I assume mix the results in the cell, maybe separated by newlines.
 - Be sure to do this in the view and export version, as applicable

sei-aderr:
> Each TestCase will have one or more QAS ids associated to it. Using that id, the frontend will have to find the corresponding QASDescriptor in the NegotiationCard data section of the report, and get the "quality" field from it. That is what should go in the cell.

Are the test results not updated to the format needed for this? After running the simple demo in the current staging branch I see a `test_case_id` field that is the "quality" field from the system requirements section: accuracy, class distribution etc.

> We should also show the QAS id in the Report. We either mix those and show everything in the same column (something like "Quality Attribute Scenario" for title, and the "qas_001 - Robustness" for a value), or create different columns for each.

Similar to above, I don't see the QAS id in the testResults that are being loaded as findings in the report page.

sei-secheverria:
Ok, I think there is some confusion there, mostly due to the legacy ids we are still using.
 - The TestResults don't have anything to say about the new format, in terms of QAS ids. The only change in the TestResults is that they are associated now to TestCases, instead of to vaguer conditions we had before.
 - The TestCases are the ones associated to the QAS in the NegotiationCard, through the qas_ids. So in the Report, you have to follow that link: from TestResult, to TestCase through test case id, to QAS in NegotiationCard through qas id.
 - It just so happens that the ids we are giving to the TestCases in the TestSuite in the demo, are words like "accuracy", "class distribution", etc, mostly because they are test cases associated with QAS focused on that.  So those are not the quality field from the system requirements section, but the name of the test ids in the TestCase.

sei-secheverria:
Addressed in #648

PR #648, opened 2025-04-28
Feature/qas-ids-report

sei-aderr:
Addresses #627 

Add support for report form and export to show QAS ID and QA in results table


sei-secheverria (approved):
Looks good!

Issue #651, opened 2025-04-29
Verify QAS Ids in TestCases


sei-secheverria:
Due to the way we are storing the QAS id in the TestCases (as a string for the id), when the test library is used to manually create either, people may make typos and indicate invalid QAS ids. Find a way to validate them, probably in the .save methods of each artifact.

Note that this issue used to contain a similar verification for QAs and NegotiationCards, which was moved over to #653

sei-secheverria:
QA/NegotiationCard check was moved over to #653. Edited this issue to remove that part.

sei-secheverria:
Addressed in #652

Issue #650, opened 2025-04-29
Add support for listing QAS/QA in the library


sei-aderr:
Add a function similar to `card.print_quality_scenarios()` from the negotiation card for the custom lists. This will allow library users to easily get a list of the QAS/QA that are contained in custom lists and use them for the library.

sei-secheverria:
Addressed in #625

sei-secheverria:
Wrong ticket closed

sei-secheverria:
Addressed in #654

PR #652, opened 2025-04-30
Feature/verify qa and qasid

sei-secheverria:
Addresses #651 by:
 - Added an explicit check before saving a TestSuite through the artifact save() method (used by the MLT test library) to see if the QAS ids associated to the test cases exist in any NegotiationCard stored in the context (model/version). If any of the QAS ids is not found in any NegotiationCard, the save fails with an exception.
 - Modified the schema for generating the QAS ids, now they have their own NegotiationCard id as a prefix as well. This allows the QAS ids to be fully unique inside a context (model/version), since we in theory support multiple NegotiationCards in a context, and the previous schema allowed for duplicated QAS ids (two QAS could have the same id, while being in different NegotiationCards). This does tie in a QAS to a NegotiationCard, so it can't be "moved" with the same id to another NegotiationCard (different versions or model/version contexts is a different issue to be handled later), but it seems reasonable for now, especially since in our current use cases we have one negotiation card per context anyway.
 - Does NOT address checking that the QA names assigned manually to a QAS in a manually created NegotiationCard are valid QAs stored in the QA CustomList. There is no easy way right now for a NegotiationCard to check this, since it does not have access to the current CustomList store. This MIGHT be possible after the changes in #650, where the CustomList store will be added to the test library's session, but it might will not be very straightforward. This was moved to #653


sei-aderr:
Simple demo negotiation card does not look to be updated and so the test suite notebook is failing.

sei-aderr:
Looks to be the same issue creating the Test Suite in the scenarios demo

sei-secheverria:
> Simple demo negotiation card does not look to be updated and so the test suite notebook is failing.

Good catch, it wasn't the negotiation card but the TestCases in the TestSuite. I'll fix that.

sei-secheverria:
Ok, should be fixed now

sei-aderr (approved):

PR #654, opened 2025-04-30
Feature/custom list entry in qas

sei-aderr:
Addresses #650 

Mostly finished, just had a couple questions on implementations

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,6 @@ docs/site
 /store*/
 pgdata/
 *.whl
+
+# TODO: Remove this check once RDBS and HTTP are implemented.
+default_store/
--- a/.gitignore
+++ b/.gitignore
@@ -49,6 +49,4 @@ docs/site
 /store*/
 pgdata/
 *.whl
-
-# TODO: Remove this check once RDBS and HTTP are implemented.
 default_store/
--- a/demo/simple/2_test_suite.ipynb
+++ b/demo/simple/2_test_suite.ipynb
@@ -73,6 +73,27 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "If we want to look at the `Quality Attribute Scenarios` or the `Quality Attributes` that are available, we can do that with these listing functions."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from mlte.custom_list.custom_list_names import CustomListName\n",
+    "from mlte.session.session import print_custom_list_entries\n",
+    "\n",
+    "print_custom_list_entries(CustomListName.QA_CATEGORIES)\n",
+    "print()\n",
+    "print_custom_list_entries(CustomListName.QUALITY_ATTRIBUTES)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
     "Now we can create our `TestSuite`, consisting of a list of `TestCases`, each of them addressing one or more Quality Attribute Scenarios from our `NegotiationCard`. When defining the `TestCase`s below, we need to set the id of the corresponding Quality Attribute Scenario we want to test in its \"quality_scenarios\" attribute."
    ]
   },
@@ -172,7 +193,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.12.8"
+   "version": "3.10.16"
   },
   "orig_nbformat": 4
  },
--- a/demo/simple/2_test_suite.ipynb
+++ b/demo/simple/2_test_suite.ipynb
@@ -85,8 +85,6 @@
     "from mlte.custom_list.custom_list_names import CustomListName\n",
     "from mlte.session.session import print_custom_list_entries\n",
     "\n",
-    "print_custom_list_entries(CustomListName.QA_CATEGORIES)\n",
-    "print()\n",
     "print_custom_list_entries(CustomListName.QUALITY_ATTRIBUTES)"
    ]
   },
--- a/mlte/frontend/nuxt-app/pages/negotiation-card.vue
+++ b/mlte/frontend/nuxt-app/pages/negotiation-card.vue
@@ -247,29 +247,6 @@ if (useRoute().query.artifactId !== undefined) {
             response._data.header.timestamp * 1000,
           ).toLocaleString("en-US");
           form.value = response._data.body;
-
-          const problemType = response._data.body.nc_data.system.problem_type;
-          if (
-            problemTypeOptions.value.find((x) => x.value === problemType)
-              ?.value !== undefined
-          ) {
-            form.value.nc_data.system.problem_type =
-              problemTypeOptions.value.find((x) => x.value === problemType)
-                ?.value;
-          }
-
-          response._data.body.nc_data.data.forEach((item) => {
-            const classification = item.classification;
-            if (
-              classificationOptions.value.find(
-                (x) => x.value === classification,
-              )?.value !== undefined
-            ) {
-              item.classification = classificationOptions.value.find(
-                (x) => x.value === classification,
-              )?.value;
-            }
-          });
         }
       },
       onResponseError({ response }) {
--- a/mlte/frontend/nuxt-app/pages/negotiation-card.vue
+++ b/mlte/frontend/nuxt-app/pages/negotiation-card.vue
@@ -211,9 +211,6 @@ const formErrors = ref({
   identifier: false,
 });
 
-const classificationOptions = useClassificationOptions();
-const problemTypeOptions = useProblemTypeOptions();
-
 const systemInformationRef = ref(null);
 const dataRef = ref(null);
 const modelRef = ref(null);
--- a/mlte/session/session.py
+++ b/mlte/session/session.py
@@ -14,7 +14,7 @@ from mlte.custom_list.custom_list_names import CustomListName
 from mlte.store.artifact.factory import create_artifact_store
 from mlte.store.artifact.store import ArtifactStore
 from mlte.store.catalog.catalog_group import CatalogStoreGroup
-from mlte.store.custom_list.factory import create_custom_list_store
+from mlte.store.custom_list.initial_custom_lists import InitialCustomLists
 from mlte.store.custom_list.store import CustomListStore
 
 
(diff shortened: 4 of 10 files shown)


mlte/session/session.py
```diff
@@ -156,3 +188,16 @@ def add_catalog_store(catalog_store_uri: str, id: str):
     """
     global g_session
     g_session._add_catalog_store(catalog_store_uri, id)
+
+
+def print_custom_list_entries(list_name: CustomListName) -> None:
```
sei-aderr, line 193: Thoughts on how we would want to adjust this so that the custom lists are always populated? The initialization of the custom lists only happens in main right now when the backend starts up so it doesn't look like these lists are guaranteed to exist when using the library.

test/session/test_session.py
```diff
@@ -33,19 +33,20 @@ def test_session() -> None:
...
 
     assert s.context.model == model
     assert s.context.version == version
-    assert s.artifact_store.uri.uri == artifact_store_uri
+    assert s.artifact_store.uri.uri == store_uri
     assert s.catalog_stores.catalogs[cat_id].uri.uri == catalog_store_uri
+    assert s.custom_list_store.uri.uri == store_uri
 
 
 @pytest.mark.parametrize("store_fixture_name", artifact_stores())
```
sei-aderr, line 52: Would I need to adjust anything here to check with the custom lists?

mlte/session/session.py
sei-secheverria: Good question. I'll add a suggestion where I think it should go.

mlte/session/session.py
```diff
@@ -84,9 +93,25 @@ def artifact_store(self) -> ArtifactStore:
...
+    @property
+    def custom_list_store(self) -> CustomListStore:
+        if self._custom_list_store is None:
+            # If the custom list store URI has not been manually set, get it from environment.
+            if self.MLTE_CUSTOM_LIST_STORE_URI_VAR in os.environ:
+                self._custom_list_store = create_custom_list_store(
+                    typing.cast(
+                        str, os.getenv(self.MLTE_CUSTOM_LIST_STORE_URI_VAR)
+                    )
+                )
```
sei-secheverria, line 107: This probably should be changed to InitialCustomLists.setup_custom_list_store(), instead of calling create_custom_list_store directly. The InitialCustomLists method does that setup, and then also the create_custom_list anyway.

mlte/session/session.py
```diff
@@ -84,9 +93,25 @@ def artifact_store(self) -> ArtifactStore:
...
         return self._artifact_store
 
+    @property
+    def custom_list_store(self) -> CustomListStore:
+        if self._custom_list_store is None:
+            # If the custom list store URI has not been manually set, get it from environment.
+            if self.MLTE_CUSTOM_LIST_STORE_URI_VAR in os.environ:
+                self._custom_list_store = create_custom_list_store(
+                    typing.cast(
+                        str, os.getenv(self.MLTE_CUSTOM_LIST_STORE_URI_VAR)
```
sei-secheverria, line 105: As a separate comment, especially since this is now done both here and in other stores in this same Session module, there may be a prettier way to handle this environment var code...having this whole check if var in os.environ, then the typing.cast, and again the URI_VAR name call has always been a bit ugly, but now it is basically duplicated here and in the other stores that do the same in this file. Maybe create a function like get_env_var(), that does the check and cast, and return None if it is doesn't exist, would be a way to avoid this duplication and make this code cleaner.

test/session/test_session.py
sei-secheverria: Yes, probably a similar check to check that the custom list is setup and has the expected uri. Similar type of changes should be done to test_environment_vars, test_no_context_setup and test_no_store_setup

demo/simple/2_test_suite.ipynb
```diff
@@ -69,6 +69,27 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from mlte.custom_list.custom_list_names import CustomListName\n",
+    "from mlte.session.session import print_custom_list_entries\n",
+    "\n",
+    "print_custom_list_entries(CustomListName.QA_CATEGORIES)\n",
```
sei-secheverria, line 88: Do we need this? The user doesn't reall care about the categories, other than as groupings of the QAs, and here we are not really using this for anything.

mlte/frontend/nuxt-app/pages/negotiation-card.vue
```diff
@@ -247,29 +244,6 @@ if (useRoute().query.artifactId !== undefined) {
...
-            const classification = item.classification;
-            if (
-              classificationOptions.value.find(
-                (x) => x.value === classification,
-              )?.value !== undefined
-            ) {
-              item.classification = classificationOptions.value.find(
-                (x) => x.value === classification,
-              )?.value;
-            }
```
sei-secheverria, line 271: Just wondering, why was this removed? It does not seem to have anything to do with custom lists.

sei-secheverria (changes requested):
Comments and answers below

mlte/frontend/nuxt-app/pages/negotiation-card.vue
sei-aderr: It is no longer used because the options for selection drop downs are handled in the shared components now and they look to work without this adjustment to the data format of the selection items.

sei-aderr: Continued in #655

test/session/test_session.py
sei-secheverria: Right, I somehow missed that it was doing that. I saw the added lines to the other tests. I think that should be enough. Or if not enough, at least it would be consistent with the previous tests.

mlte/session/session.py
sei-aderr: Looking into this and other comment now. `create_custom_list_store` accepts `uri: str` where `setup_custom_list_store` accepts `store_uri: StoreURI` so trying to figure out what the best way to handle that is. Let me know if you have any strong opinions on how it should be handled. Current easiest solution I think would just be parse the URI in session and then pass it along.

sei-secheverria: Right, I noticed that too. As you say, probably just calling StoreURI.from_string(uri) before passing it along would be enough.

But another option would be to take this opportunity to change this, both for setup_custom_list_store and for setup_sample_catalog. I see no good reason, from their structure or from where they are used, that they really need to receive a parsed StoreURI. We could change both methods to receive a uri str instead, and do the parsing inside. That would make this two pseudo-factory methods much more consistent with the other factory creator methods we have.

demo/simple/2_test_suite.ipynb
sei-aderr: Agree that it seems out of place, but I thought we had agreed we wanted to show the functionality in the demos? I am fine moving it somewhere else but wasn't sure where else it would go.

mlte/session/session.py
sei-aderr: I did actually just go that route with the the initial custom lists setup. If the implementation looks good and everything works I can adjust the catalog one as well.

demo/simple/2_test_suite.ipynb
sebastian-echeverria: Right, my point being print_custom_list_entries for Quality Attributes should be enough to show that functionality, and to show the QAs, which is all the user cares about at this point. Not sure we show any new functionality by showing the QA Categories, and the users don't need them separately from listing the QAs to know which QAs to put in their QASDescriptors.

.gitignore
```diff
@@ -49,3 +49,6 @@ docs/site
 /store*/
 pgdata/
 *.whl
+
+# TODO: Remove this check once RDBS and HTTP are implemented.
+default_store/
```
sei-secheverria, line 54: We will not really remove this. The sample catalog does always default to FS if a non-FS store is used, it is its own special behavior. I just changed the name of the default folder to make it clearer, and currently it is being shared by the custom list RDBS/HTTP workaround, but once we remove that, this will stil stay for the sample catalog. I'd remove this particular comment.

mlte/store/catalog/sample_catalog.py
```diff
@@ -41,19 +41,22 @@ def setup_sample_catalog(
         # Set file system URI if we didn't get one, or if we got a non-file system one.
         if stores_uri is None or (
             stores_uri is not None
-            and stores_uri.type != StoreType.LOCAL_FILESYSTEM
+            and StoreURI.from_string(stores_uri).type
+            != StoreType.LOCAL_FILESYSTEM
```
sei-secheverria, line 45: I get this, but this is duplicating the from_string called below in the else. Unless you have a reason for this, I'd recommend moving the parsed_uri creation currently in line 51 to line 41, before the if, and reverting this check back to how it was (well, with parsed_uri instead of store_uri). This would also allow you to remove the else. Having the else is slightly clearer, but the fact that we have to call from_string twice and parse the uri twice, in my opinion, removes that advantage.

mlte/session/session.py
```diff
@@ -156,3 +187,16 @@ def add_catalog_store(catalog_store_uri: str, id: str):
...
+    """Prints custom list entries in a user-friendly way."""
+    global g_session
+    entry_list = g_session.custom_list_store.session().custom_list_entry_mapper.list_details(
+        list_name
+    )
+    for entry in entry_list:
+        if entry.parent:
+            print(f"{entry.name} (Parent: {entry.parent}): {entry.description}")
+        else:
+            print(f"{entry.name} (Parent: None): {entry.description}")
```
sei-secheverria, line 202: Just wondering, could this be encapsulated in the __str__() method of a CustomListEntryModel?

sei-secheverria (changes requested):
Very good, I have only a few, very minor, nitpicky suggestions, sorry!

mlte/store/catalog/sample_catalog.py
sei-aderr: I agree that I don't like that it is duplicated and tried a couple different ways. Moving the parsed_uri creation to line 41 will not work because the `stores_uri` param being passed to `setup_sample_catalog` is optional and `StoreURI.from_string` only accepts a string. I tried several different ways of setting it up but couldn't get one to be accepted by type checking.

I can look at it again and see if there is a different way to order things to only have to call it once or if you see another way.

mlte/session/session.py
sei-aderr: Good idea, added

mlte/store/catalog/sample_catalog.py
sei-secheverria: Very good point. However, looking at it and playing with it, I realized the Optional hint for the uri was not needed in this function, as it is always called with the uri string. This was probably an artifact of a previous version of this function that was never fixed. Could you try doing this change and removing that Optional = None option from the function def? It should work in that case.

sei-aderr: If that is not needed then it is much simpler. Pushed the change

sei-secheverria (approved):
Looks good!

Issue #605, opened 2025-03-11
Consider Updates to the MLTE PyPI


krmaffey:
Consider if it's worthwhile to switch MLTE's PyPI presence from MLTE-python to MLTE.

krmaffey:
After reviewing several articles and and a GitHub repository offering a transition helper template, it seems like the best course of action is to follow the steps laid out in the Python documentation ([link](https://peps.python.org/pep-0423/#how-to-rename-a-project) to overview section).

Here are my explanations of the steps along with links to portions of the documentation I found helpful.

1. Publish the package under the new package name (mlte on [PyPI](https://pypi.org/manage/project/mlte/collaboration/)), ensuring package files are updated with the new name as appropriate
     - When publishing the new distribution, assign obsolete metadata in the `setup.cfg` file as per this [documentation](https://packaging.python.org/en/latest/specifications/core-metadata/#obsoletes-dist-multiple-use)

2. Edit the old package (mlte-python on [PyPI](https://pypi.org/manage/project/mlte-python/releases/)) and release it one more time with two additions to the `setup.py` file (an example can be viewed [here](https://github.com/simonw/pypi-rename-demo/blob/main/setup.py))
    - Add the new package as a dependency
    - Add the `Development Status :: 7 - Inactive` classifier in setup script

These steps should ensure that users are directed to the correct version of `MLTE` despite the name change. Please let me know what questions you have @sei-secheverria / @sebastian-echeverria !

sei-secheverria:
According to this, https://packaging.python.org/en/latest/specifications/core-metadata/#obsoletes-dist-multiple-use, obsoletes-dir is a very rarely used metadata field, so even though it is recommended by the renaming docs, is unlikely anybody is using it, incuding pip or PyPi, for anyhing. In particular, according to this (https://github.com/python-poetry/poetry-core/blob/main/src/poetry/core/masonry/metadata.py), Poetry accepts the obsoletes-dist metadata field, but it does not set it to be stored inside the packaged info, so, for example, it does not show up inside the PKG-INFO inside the sdist.

It seems as if it is not that useful right now, but it does not really hurt, so we'll leave it in the metadata, at least it could help someone looking at the pyprorject.toml

krmaffey:
[...]

PR #658, opened 2025-05-02
Updates to v2.0 for new PyPI name

krmaffey:
Addresses step 1 in #605, preparation for publishing v2.0 of MLTE via our new name on PyPI: mlte (old name was mlte-python).


pyproject.toml
sei-secheverria, line 1: Everything looks ok. As described in #605, poetry (and mostly anybody else) seem to ignore the obsoletes-dir metadata field, but we'll leave it for consistency.

sei-secheverria (approved):
Looks good!

Issue #32, opened 2022-04-28
Fluent Mechanism for Specifying Units


turingcompl33t:
Many of the implementations of `Measurement`s require specification of some threshold value. We should develop a way to make the units explicit in the API of these functions.

sei-secheverria:
One option for supporting this would be to add a unit argument to every method that creates a Validator. This, however, would be hard to enforce and structure. Regardless, if we did this, the Validator should also store the unit that is being used.

Another option would be to create some sort of class to represent a unit-annotated threshold, say a Unitary class that consists of a primitive plus a unit. Thus, in every method that creates a Validator, one of this objects would be passed. This could look like this:

`
Real.greater_or_equal_to(Unitary(0.98, Unit.percent))
`
Or:
`
LocalObjectSize.get_output_type().less_than(Unitary(3000, Unit.bytes)),
`

Units could be a enum with defined values, that should be expandable, or just a string indicating the unit, which may be valid since we will not really use it, but use it as documentation.

Regardless, storing this explicitly in the Validator would be useful for traceability and auditing. 

Alternatively, we could somehoe reuse the Evidence class and derivatives, and store units in them, using them not only for Evidence but also for thresholds. This could be confusing with their current name, though.

sei-secheverria:
Another, maybe complimentary option, would be to use an existing unit library. Some options:
 - metpy (https://unidata.github.io/MetPy/latest/index.html) or astropy.units (https://docs.astropy.org/en/stable/units/index.html), which are focused on meteorology and astronomy, but have lots of units; however, these seem more focused on conversion between units
 - python-measurement (https://python-measurement.readthedocs.io/en/latest/) and quantities(https://github.com/python-quantities/python-quantities), or pint (https://pint.readthedocs.io/en/stable/getting/overview.html), which again, seems focused on conversion between units, but could be used to be unit-aware

sei-secheverria:
Addressed in #662

PR #662, opened 2025-05-13
Feature/units

sei-secheverria:
Addresses #32 to add support for units in measurements and evidence. More specifically:

 - PR adds integration with pint library to handle units. Pint is not accessed directly though, measurement.unit module is created as a wrapper/facade to simplify access to library.
 - Unit module provides a Units object, that has attributes defining most common units, as well as a Unit and Quantity type (Quantity contains a magnitude and a Unit).
 - Integer and Real Evidence types were updated to optionally support Units, if needed. Units can be defined for a new Integer or Real being created, as well as passed as units to thresholds associated to methods that create Validators. Both classes were updated internally to handle units properly in their methods that define bool expressions when creating Validators.
 - Validator was updated to record thresholds with their units for auditing purposes.
 - CpuStats and MemoryStats objects were updated to record the unit they are storing the data in explicitly.
 - Cpu, Memory and Storage Measurements were updated to 1) add the corresponding unit to the Evidences they generate, 2) optionally receive a target unit when evaluating, to convert their internal results to that output unit, and 3) internally handle units properly in methods that create bool expressions when creating Validators.
 - ExternalMeasurement was updated to allow external functions to return units along with their results, and pass them on transparently to the output Evidence type the ExternalMeasurement evaluation method returns.
 - Demos were updated to use units where appropriate.
 - Some unit tests were fixed, and some minor inconsistencies in the demos were fixed as well.


sei-secheverria:
> So it looks like these changes populate units into the creation of the `TestCase`s, which then looks like it could be accessed in `TestResults` through the `test_case_id`. This part is alright, but how do you see this translating to display on the frontend or working with the negotiation card?
> 
> I suppose the argument could be made that it doesn't need to, and that you just need to specify them when you are writing the tests and when you do that, you could reference the text fields in the negotiation card that could specify units along with the values. Then just have to find a way to display them in the report. Is that your intent?

Very good questions. This PR is adding Unit support to basically two underlying entities: Evidence and Measurements. In practice, this means it is adding unit support to Integer and Real Evidence types, to the three built-in Measurements we have, and to the two corresponding measurement-specific Evidence types, plus a minor change to ExternalEvidence to ensure units work there as well, and minor changes to Validators (mostly to methods that create built-int Validators) to handle units if present. 

This means that the NegotiationCard is not affected at all, as none of these entities are included in the NegotiationCard. The NegotiationCard can have unit-related text in its Response Measure field, but that is just text, so it is not involved in this update, nor can it be involved. But the Negotiation Card QAS will be the source of the units that people should use later, so conceptually it will be the source people will use when deciding the units to put later. So there doesn't need to be any changes on how the NegotiationCard is handled/shown, nor how the NegotiatioCard section in the Report is shown. This PR is to allow TestCases to handle units properly, between Validators and Measurements, especially for annotating TestCases more clearly, but also to take into account the actual units when validating.

TestCases in the TestSuite would be affected, as now you can indicate the unit for the things you are measuring. You should define units in your TestCases' Validators consistently with the ones used by default or defined explicitly in your Measurements, so that the validation itself works, if you are using unit-aware Measurements/Evidence types. This does mean that if you are using a Measurement that handles dimension-aware things with units, you have to either ensure you are passing it or to the validator things in their default unit, or setting the unit explicitly. This is mainly important when 1) defining a new measurement, or a unit-aware ExternalMeasurement; 2) defining new Validators for unit-aware values, and 3) when defining the threshold when creating a Validator in the TestCase, and when passing inputs if using the run_measurements option of a TestSuite, for tests that are unit-aware.

But from the point of view of the TestResults and the Reports, it is unlikely units would be handled directly. Both the TestResults and Report just show the results of a validation, which means that, by that time, all unit comparisons have been made. Any unit information will end up as Validation messages (like "Failure: bla meters is les than blabla meters").  The Report doesn't print out Evidence directly (and even if it did, it would be a call to str(), which means it does not have to know about units either, as that is handled by str()). 

Also note that you could in the NegotiationCard say "memory has to be less than 60 MB" and then in the corresponding TestCase say ".less_than(60000, Unit.kilobyte)", and that works, as they are functionally the same thing, and all measurements and validators understand and convert between units properly, using pint underneath.

So no change is needed in the UI, and almost no effect will happen there, other than validation messages in the report now will sometimes contain units in them, when the values are not dimensionless and you used units in your corresponding validators and measurements.

sei-aderr (approved):

Issue #570, opened 2025-01-10
Integrate frontend lint/typecheking into makefile and GithubActions workflow


sei-secheverria:
We have commands for linting and typechecking the frontend, but we are not using them, and they are not run with the make and GithubActions flow commands to check the code. Integrate them, as well as fix all current errors from those checks.

sei-aderr:
`eslint-config-standard` no longer looks to be needed as a separate package. Using `eslint.configs.recommended` that comes by default as well as `tseslint.configs.recommended` which are the default 2 that are suggested by typescript-eslint.

Removed `eslint-plugin-n` and `eslint-plugin-promise`. `n` is for node, so likely added by mistake. `promise` relates to javascript promises which are currently not used anywhere in the codebase.

sei-aderr:
Have found a stable setup for linting. Nuxt ESLint claims to install JS, TS and Vue checks so that they would not explicitly be needed, but had some issues with that. The Vue eslint I had installed was able to be removed easily and still gave the correct errors so that explicit dependency was removed. Because the lint we have setup used some of the Typescript parser, and other library objects trying to remove that caused several issues and so that explicit dependency  was left in and I also left the explicit declaration of the rulesets that we are using at the top of the eslint declaration to make it clear what is being used from typepscript, and because it just actually worked with that setup. Vue is not explicitly there, but the `withNuxt` declaration does imply that it is.

PR #664, opened 2025-05-21
Add frontend linting to CI

sei-aderr:
Partially addresses #570

Update ESLint to version 9 and ESLint config to the new flat format so that ours is no longer deprecated. Frontend linting is now ran in the CI pipeline. Lots of work done to attempt to simplify our ESLint setup and our dependency list to the best of my ability. Some issues were encountered and are discussed in the #570 thread, but the end result is a working linter.

Due to the large amount of Typechecking issues that were thrown by `vue-tsc`, those will be implemented in a separate PR so that this initial working lint is merged and able to be reviewed independently and easier.


sei-secheverria (changes requested):
Everything here looks good, but a couple of minor things are missing:
 - The makefile should be modified to that both qa and check-qa call the corresponding frontend lint stuff (is there a check mode for the frontend linter, or does it always fix stuff? or does it never fix stuff?)
 - The makefile also has a typecheck-frontend command that I don't see in the ci.yaml file. Did you check/ensure that this works? At least it needs to be added to the CI flow.

sei-secheverria (approved):
Approved

Issue #525, opened 2024-11-15
Make Report permanent snapshot


sebastian-echeverria:
We want the Report to be a snapshot of a test run in time. It should not change if other artifact changes. Currently it depends on ValidatedSpec/TestResults, which in turn depends on Spec/TestSpec. If those change, the current Report data will change accordingly, which would violate having it being a snapshot.

Modify Report so that:
 - It contains a standalone or permanent version of the content of NegotiationCard, TestSpec and TestResults. One way of doing this is by having the Report contain a copy of all the data in each of those artifacts. However, since #520 will be adding versioning support to NegotiationCards and TestSpecs, that means that each iteration of a NegotiationCard and TestSpec will never be lost/changed, as each change will create a new version. If this is so, it may be enough for the Report to link to the specific version of the NegotiationCard and TestSpec that was the current one when it was created/results where run. However, for TestResults, they will not be versioned since they are just intermediate results, so they will have to somehow be copied/maintained for the Report.
- A sort of versioning will be needed for Reports, by timestamp. Each time a new Report is created, it will have to be stored and clearly identified by its timestamp. However, it may be possible to modify specific fields in it (TBD).

sei-secheverria:
Addressed in #671

PR #671, opened 2025-06-06
Feature/report snapshot

sei-secheverria:
Addresses #525. More specifically:
 - Restructures report to now contain a full copy of not only the NegotiationCard, but also the TestSuite and TestResults that were used to build that report. This makes a report now self-contained, regardless of how other artifacts change.
 - Reports will now be implicitly treated as read-only and timestamped. When save()ing a report, its defined id will be updated to append a timestamp at its end, before it is stored. That means that a store saving a report will treat each save as a new artifact, achieving the expected result.
 - NegotiationCard's structure was simplified, removing the nested nc_card structure, and leaving everything at the top level. (This was only used before to share that data with the Report, but with the new structure it is not needed).
 - Minor simplification to artifact's default ids.
 - Artifact RBDS Store implementation was refactored to match with changes above to Report and other artifacts, as well as to better isolate and name the modules that handle metadata and factory of the different artifact types. Unneeded code was removed as well.
 - Model for Artifacts in RDBS Store was changed slightly, with the DBArtifactHeader table converted into a DBArtifact table, which will be the main table used when searching artifacts. The other, artifact-specific tables will basically contain the body, and it is possible to have bodies with no associated entry in DBArtifact, which is used to store the embedded bodies of the NegotiationCard, TestSuite and TestResults for a Report.
 - Frontend was updated to work the same as before with the changes to NegotiationCard and Report.
 - Frontend bug that was not showing results in the Edit view of Report was fixed.
 - Frontend dependencies were pinned and reorganized better between dev and non-dev ones.
 - Demos and sample NegotiationCard were updated to reflect the changes above.
 - Several bugs in unit tests were fixed, and all corresponding tests were also updated as needed.


mlte/report/model.py
```diff
@@ -32,11 +34,23 @@ class ReportModel(BaseModel):
     artifact_type: Literal[ArtifactType.REPORT] = ArtifactType.REPORT
     """Union discriminator."""
 
-    nc_data: NegotiationCardDataModel = NegotiationCardDataModel()
+    negotiation_card_id: str
```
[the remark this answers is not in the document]
sei-secheverria: Good point. Because of how artifacts are structured, we don't store everything about a negotiation card in the report; we store everything except for the header. Some header info is not important here, like creation time or username that created it, but the identifier is also only in the header. Since in theory we currently support multiple negotiation cards, and we want to add support for traceability among different versions of a model as well, the main goal for storing this here is to indicate what was id of the negotiation card being used, so that, if we want a view where we can track changes and such, we don't lose this reference. We may not end up using it, though, but I'd rather keep that information with our current structure.

sei-secheverria:
New changes look good to me. If you have finished reviewing on your part we should be done, but we'll hold up merging this until we get 2.1.0 out.

sei-secheverria:
> Seeing this error in the frontend when I load into any page. Seems a package may have an issue in the new dependency structure but don't see anything actually breaking.

Are you still seeing this error? I have not been able to reproduce, at least in the Docker environment.

mlte/report/artifact.py
```diff
@@ -40,29 +37,33 @@ class Report(Artifact):
...
+        test_suite: Optional[TestSuite] = None,
+        test_results: Optional[TestResults] = None,
         comments: List[CommentDescriptor] = [],
         quantitative_analysis: QuantitiveAnalysisDescriptor = QuantitiveAnalysisDescriptor(),
     ) -> None:
+        """
+        Creates a Report.
+
+        :param identifier: The Report id (default value used if not provided).
+        :param negotiation_card: A NegotiationCard object; if None, NegotiationCard with default id will be loaded.
```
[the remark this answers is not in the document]
sei-secheverria: The current workflow we have, which may change, is that you have one negotiation card per model version. You should also have one negotiation card per multiple versions, but that is a separate issue. Even though we support multiple negotiation cards, we don't really have a workflow we recommend when that would be used. That means that, the most common case would be to have one negotiation card per model/version. And if that is the case, there is no good need to give the negotiation card an id other than its default id. So what I see is that the most common case would be for the user to have only one NegotiationCard, using the default id. Thus, when creating a Report, by default you will want it to load the default NegotiationCard and use it, simplifying what most users will have to write when creating the report. The only potential cause for the user not to commonly use the default id would be that the UI encourages the user to give the negotiation card a unique id, but that is something I think we should change at some point, and suggest the default id and just allow the user to change it if needed. But even in this case, when creating the NegotiationCard manually, it is still likely the user will use the default it. Thus, loading the default negotiation card by default in the report makes sense. If the NegotiationCard being used does not have the default id, not providing it when creating the Report will result in an error, which will just tell the user that they have to provide the actual id, if they are using a user-created one. 

At some point, we may actually restrict having more than one negotiation card per model/version, and having the default option of not providing a card id when building the report would make even more sense then. Part of one way of doing this in the future is to simply force the card to have only the default id, preventing the creation of new, separate cards.

This kinda extends to the other cases. It is likely people will have one testsuite per model version; and even if they sometimes don't, the simpler scenarios will have only one TestSuite, as we do with our current demos, where there is no need to use an id other than the default. It should be even clearer with the TestResults, since those are automatically generated by the validator, so unless you manually go out of your way and change its default id before changing, they will have the default id.

The goal of all of this is to make it simpler for people creating the Report, to simply use the (most likely) one card, one suite, and one results that they have in their model/version anyway, without having to deal with ids that won't really identify them from any other one, since they will only have one of each.

Makefile
```diff
@@ -96,6 +96,10 @@ frontend-env:
 	npm install && \
 	npx gulp init
 
+.PHONY: frontend-env-clean
+frontend-env-clean:
+	rm -rf mlte/frontend/nuxt-app/node_modules
```
sei-aderr, line 101: If we want this to be a full clean it should also remove `nuxt-app/.nuxt`, `nuxt-app/.output`, and potentially `nuxt-app/assets/uswds`

mlte/store/artifact/underlying/rdbs/suite_factory.py
```diff
@@ -0,0 +1,75 @@
...
+        test_suite_orm.test_cases.append(test_case_orm)
+
+    return test_suite_orm
+
+
+def create_suite_model(
+    test_suite_orm: DBTestSuite,
+) -> TestSuiteModel:
+    """Creates the internal model object from the corresponding DB object."""
+    # Creating a TestSuite from DB data.
```
sei-aderr, line 52: Second comment feels uneccessary

mlte/report/artifact.py
```diff
@@ -136,11 +106,33 @@ def from_model(cls, model: BaseModel) -> Report:
         body = typing.cast(ReportModel, model.body)
         return Report(
             identifier=model.header.identifier,
-            system=body.nc_data.system,
-            data=body.nc_data.data,
-            model=body.nc_data.model,
-            system_requirements=body.nc_data.system_requirements,
-            test_results_id=body.test_results_id,
+            negotiation_card=NegotiationCard.from_model(
```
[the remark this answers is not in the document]
sei-secheverria: Before it was copying each part of the negotiation card model into the Report constructor arguments. However, now we are embedding the whole card, suite and results as whole Artifacts, to avoid having duplication for each field in each artifact. The reason of the use of from_model, was to try to reuse the initialization for each artifact type, since for the three of them, we have the model and we are converting it to a (sub or internal) artifact in the Report artifact. One reason it is too verbose is that we actually don't have the full artifact model for the card, suite and result in the report model, but only the body. So we are recreating the artifact model by adding the type and id, plus the body, into a fulll artifact model, and then creating each sub-artifact from there. There may be a way to make it cleaner, though. I would have to check how the Report is using the artifacts. One option would be simply to store the models inside the Report. I am not totally sure we need or want the full artifact inside the Report artifact. The main goal of the Artifact structures is to do manipulation on each artifact, which may not be needed for this embedded artifacts.

Makefile
sei-secheverria: Done

Makefile
```diff
@@ -96,6 +96,13 @@ frontend-env:
...
+	rm -rf mlte/frontend/nuxt-app/.nuxt
+	rm -rf mlte/frontend/nuxt-app/.output
+	rm -rf mlte/frontend/nuxt-app/assers/uswds
```
sei-aderr, line 104: assets typo

sei-aderr (dismissed):

Makefile
sei-secheverria: Fixed

sei-aderr (approved):

Issue #674, opened 2025-06-11
Bug: crash when SQLAlchemy is not installed


sei-secheverria:
MLTE crashes when the optional SQLAlchemy lib is not installed. This dependency is supposed to be optional, and is not installed by default, only manually if needed. However, due to imports in the factories, it is always loaded, leading to crashes. Modify so that it is only imported if really used in the store factories.

sei-secheverria:
Addressed in #675

PR #675, opened 2025-06-11
Bugfix/sql alchemy crash

sei-secheverria:
Addresses #674 by moving imports to the corresponding RDBS implementations for the different stores, to the case in each factory where the RDBS is instanced, to avoid trying to load SQL-related libraries when they have not been installed, since they are now optional and not installed by default. 

High priority fix, since currently released 2.0.0 version will crash every time if SQL libraries are not installed.

Other minor fixes:
 - Simplified makefile command for black formatter adding Jupyter files to its configuration
 - Added poetry lock (without updates) to venv command to always ensure it is up to date before installing
 - Minor removal of (now) unneeded cast in simple demo.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,7 @@ venv-clean:
 .PHONY: venv
 venv:
 	python -m venv .venv && \
+	poetry lock && \
 	poetry install --with dev,demo --all-extras
 
 .PHONY: venv-redo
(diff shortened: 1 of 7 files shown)


sei-aderr:
Under what conditions were you seeing a failure on master to cause this issue? Currently on master if I do a new venv with `poetry install`, I get an error that `uvicorn` isn't installed, not anything related to sql

Issue #513, opened 2024-11-15
Handle Human Input in Report


sebastian-echeverria:
There are two cases where we may need human input to end up in the report:

- When either creating the report or running the validation before it, a human should be able to add its validation through code, right before the Report is generated. This would apply to TestCases where the validation (whether it succeeded or not) can't be done automatically (such as for Interpretability).
- Would it be needed for humans to add additional input to the Report? Or does the case above cover the main reason? Do we want to allow humans to add "general comments"?
- Evaluate how this interacts with #525 , especially the manual comments

sei-secheverria:
Addressed in #683

PR #683, opened 2025-06-21
Feature/human validation

sei-secheverria:
Addresses #513 by adding a method to TestResults that allows you to manually convert an Info result into Success/Failure (after supposedly manual revision).  Updated simple demo to show how this would be use. This would probably need some validation from the team before being complete, plus some discussion about comments and quantitative analysis fields in report.


sei-secheverria:
> I now see that you mentioned that too in the PR, I think we remove the section for sure. If they need to be revise the results in the library before creating the report I don't see any need to have the section in the frontend.

Right, we should remove that. Go ahead and remove it when you can please.

mlte/evidence/types/image.py
```diff
@@ -91,6 +91,12 @@ def register_info(cls, info: str) -> Validator:
         )
         return validator
 
+    # Overriden.
```
sei-aderr, line 94: Were these just all needed to do the additional validation in their actual type? So you want to be able to load the image as an image, and not as an evidence so it can be displayed correctly? 

What would for example, the array have looked like if it was loaded as an evidence before?

sei-aderr:
> > I now see that you mentioned that too in the PR, I think we remove the section for sure. If they need to be revise the results in the library before creating the report I don't see any need to have the section in the frontend.
> 
> Right, we should remove that. Go ahead and remove it when you can please.

Got the section removed, do we want to leave in the heading `Test Results (Quantitative Analysis)` or shorten it to just `Test Results`?

sei-secheverria:
> > > I now see that you mentioned that too in the PR, I think we remove the section for sure. If they need to be revise the results in the library before creating the report I don't see any need to have the section in the frontend.
> > 
> > 
> > Right, we should remove that. Go ahead and remove it when you can please.
> 
> Got the section removed, do we want to leave in the heading `Test Results (Quantitative Analysis)` or shorten it to just `Test Results`?

Let's shorten it.

Issue #678, opened 2025-06-16
Add docstrings to functions in frontend


sei-aderr:
Most functions in the frontend are not documented. Add at least simple docstrings for simple functions, and full explanations for more complex/shared ones

Issue #672, opened 2025-06-11
Look into standardization of api call error handling


sei-aderr:
Typechecking PR is going to introduce typing to all the api endpoints so some of the `useFetch` that were more complicated like in the negotiation card are no longer going to have the `onResponse` section. So the check for `isValidNegotiation` is outside of the `useFetch` which should be fine because `onReponse` should have been called every time before. 

Problem is that the error checking is weird now. Planning to change `if (isValidNegotiation(cardData)) {` to `if (cardData && isValidNegotiation(cardData)) {` so that it won't run if there isn't card data but then the page still loads a blank page. So all more complicated `useFetch` will be adjusted to that for now but even that isn't the best. Then several other of the api calls don't have error checking on them for things like `QACategoryAPIData`.

Look into standardizing this in some way and ensuring that failure cases are handled properly.

sei-aderr:
Addressed in #689

PR #689, opened 2025-07-21
Refactor API Calls

sei-aderr:
PR 2, review after #687 

Addresses #672, partially #678

Finally found a way to create a generic API handler that can be typed correctly. All frontend API calls are adjusted to use this new format.

### DIFF
--- a/mlte/frontend/nuxt-app/components/Admin/user-edit.vue
+++ b/mlte/frontend/nuxt-app/components/Admin/user-edit.vue
@@ -199,7 +199,6 @@ async function submit() {
       inputError = true;
     }
 
-    // TODO : Bug fix this between edit and new user style
     if (
       props.modelValue.password == undefined ||
       props.modelValue.password.trim() === ""


sei-secheverria:
So much better. This is exactly what we needed to remove some of that duplication and simplify the API access.

In terms of creating wrapper functions for each API access, there are some pros and cons there. One one hand, this would centralize all URLs, so that we don't have URL strings for specific API parts all over the place as now, and it would also reduce duplication (for example, useApi("/user/me") is reused in at least a couple of places), allowing us to make it easier to change URL paths without having to go all over the code for this. It would also make intent a bit clearer (getUser is easier to read than useApi("/user", GET), but not by that much). On the other hand, we may end up with a lot of wrapper functions. But back to the pros, that would add a clean abstraction all over the code, with only the api.ts file having to know not only the strings for the URLs, but how each API call is actually formed. From a design perspective, this does sound cleaner, as you are adding a layer of abstraction for the API so that all the rest of the code does not need to know how the API actually works in terms of URL structure, HTTP methods, etc. The useApi method is already one big step in that direction, the wrappers would be the final step to totally abstract that out from the rest of the code. We can discuss further if you want.

sei-aderr:
Sounds like you are heavily leaning towards adding wrapper functions so I am happy to go that route. It will definitely be quite a few wrapper functions with a lot only being used in one place, but some will be used a couple times. And agree, it does make adjusting any calls to the api a lot easier to adjust, assuming we can always end up with the same data that the page is using.

I had gone the route of adding them for the needed endpoints in #692, but I will add all the ones that could be implemented in this PR, and then figure out a merge to the next PR after its finished.

sei-secheverria (approved):
Looks good!

Issue #568, opened 2025-01-10
Have a View option for Artifacts


sei-secheverria:
Even though #481 has something like this for the test catalog, it would be useful to have just a View option for all artifacts as well. It could be the same Edit view with changes and buttons disabled as well, or something different.

sei-secheverria:
Finished addressing in #692

PR #692, opened 2025-07-25
Feature/artifact view

sei-aderr:
PR 3

Addresses #568


sei-aderr:
The opaque and array value types have data typed as any. Typescript typechecker doesn't like things being typed as any. Ist here a subset of types we can expect them to be? Could just ignore the error or look into other options if it really is an any

Issue #660, opened 2025-05-12
Show Evolution of Report Results Between Model Versions


sei-secheverria:
Currently, results in reports are totally independent between reports. However, it would be useful to see how TestCases evolve through versions, as well as how QAS are either handled successfully or not. Create a view in the Frontend UI that allows people to see this evolution. This may also be useful among different versions of a report for the same model version.

Note that this depends on TestCases and QAS in NegotiationCards not changing between model versions, or at least not being reused or changed to mean something else. Part of this can be addressed simply by assuming the user will not be abusing this (not changing the contents of a QAS/TestCase to something else entirely, instead of deleting it and creating another), and in theory, new or deleted QAS or TestCases and simply be shown as not having a previous or further evolution. However, this has to be made somehow clear.

sei-secheverria:
I am leaning towards yes, having one TestSuite per model/version. The closest you can do for now is to assume the default id for the TestSuite, potentially allowing that to be changed if you want.

sei-secheverria:
Addressed in #700

PR #700, opened 2025-08-13
Feature/artifact-compare

sei-aderr:
Addresses #660 

Initial implementation of a compare page for reports and test cases. Will see how it looks initially and make adjustments

### DIFF
--- a/mlte/frontend/nuxt-app/components/report-compare.vue
+++ b/mlte/frontend/nuxt-app/components/report-compare.vue
@@ -1,6 +1,11 @@
 <template>
   <h3>Header</h3>
-  <div><b>Differences: </b> {{ headerDifferences.join(", ") }}</div>
+  <div v-if="headerDifferences[0] == 'None'" style="color: green">
+    <b>Differences: </b> {{ headerDifferences.join(", ") }}
+  </div>
+  <div v-else style="color: red">
+    <b>Differences: </b> {{ headerDifferences.join(", ") }}
+  </div>
   <b>{{ version1 }}</b>
   <HeaderDisplay :header="report1.header" />
   <b>{{ version2 }}</b>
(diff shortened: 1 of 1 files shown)


sei-secheverria (changes requested):
Changes work great. A couple of things we could add/improve:

 - The names of the reports tend to be long due to the timestamp being added to them, but the current HTML input field tends to be smaller than the default report names. I'd increase the size of the HTML dropdown input field for selecting which report to use, so for the default report name, the full name/timestamp can be seen after it has been selected.
 - When there are no differences in a test case results, it shows "Differences: None". But for the header, it just shows "Differences: ", with just a blank space, which is confusing as it seems to show the differences are in the lines below. We could add the "None" word for that case in the header as well.
 - Even though the "Differences" section is very useful to highlight differences, it is not easy to quickly skim and find the cases that have differences. I'd recommend using colors for this. Maybe the simples would be to use a background color of green for "Differences: None", and a background color of red or similar for "Differences: <something not None>". This would make it much easier to quickly skim and focus on the differences.
 - A bit more complex, and maybe not necessary, it could be useful to also highlight in some color the fields described in the list of Differences. So if the differences say "Differences: Status, Message", then highlighting the status and message fields of the two things being compared. However, this may be tricky, so if it is not relatively trivial, we can skip this, since it is not that critical.

sei-aderr:
Addresses first two comments.

I agree about it being not super clear which things have differences and don't and had thought about options but hadn't implemented anything yet. 

For your first idea, should be a fairly simple implementation, but not sure how good it will look with the lack of large pieces of color in the rest of the page. Just super simple test, it could look something like this. Makes it clear what is different, but is a bit blunt
image

Second idea, probably wouldn't be impossible to implement. I think we would have to pick between that and the first option though so we only have one thing being highlighted in color.

I was thinking something more like this, where the differences text is color coded to whether or not there is a difference. From just initial test, the red is pretty clear to find, but the green blends in a bit. Which isn't strictly bad because red is what you would want to be looking at anyway.
image

sei-secheverria:
The second image looks perfect. I think for that is enough. Let's ignore comment 4 about coloring the change. I think, for now, coloring the text that talks about the differences is enough.

Issue #661, opened 2025-05-12
View to Track Evolution of QA Scenario Results


sei-secheverria:
Similar to #660, it may be useful to have a view that tracks the evolution per QAS, and also maybe by TestCase. This would be another way to view #660, and it may potentially replace it. The idea is to be able to visually track how results by requirements have evolved, between runs and model versions.

sei-aderr:
Did you have an idea how you thought the QAS would be displayed? I was starting to work on it, with the plan to have it work like the table in the report view then realized that doesn't work because we are showing all of the test cases in the one row. With the space constraints we already have, I don't think it is reasonable to hope that they could end up within the table directly. I think best case would be work it in as an on-hover or modal for each result similar to how we want to have an expanded view of the results available but I'll see what we end up looking like

sei-aderr:
- Updated the table to be color coded similar to the results table. 
- Adjusted column headers to grab any `test_result_id` from any report in the model

Shortening the `Success/Failure` values didn't shrink the table at all because the column headers took up more width. So adjusted the table to side scroll if it runs out of horizontal space

Because I can't directly use the `UsaTable` component when manipulating the color and style of the table cells, we unfortunately lose the easy implementation of sorting on certain columns. I will have to look into that more later to see if I can manipulate the component somehow, or may have to implement our own custom sort functionality.

sei-secheverria:
Ok, fair enough. I'd say sorting is less critical, and could be moved into a separate ticker for later.

PR #703, opened 2025-09-04
Feature/report evolution

sei-aderr:
Addresses #661

### DIFF
--- a/mlte/frontend/nuxt-app/assets/css/styles.css
+++ b/mlte/frontend/nuxt-app/assets/css/styles.css
@@ -90,7 +90,9 @@ hr {
 }
 
 .tooltip {
-  position: absolute;
+  position: relative;
+  display: inline-block;
+  border-bottom: 1px dotted black;
 }
 
 .tooltip .tooltiptext {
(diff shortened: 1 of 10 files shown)


sei-aderr:
Yeah, the tooltip CSS was not cooperating within the scrollable div for the table. I adjusted it to work nicely when it wasn't scrolling, but looks like it really broke when the scroll was there.

I pushed a change which was the initial version I had that I didn't like. Because the tooltips are contained within the scrollable div holding the table, it will expand the div with scrolls to fit the tooltips and they won't display outside of that which is not ideal. Can't figure out a solution that makes it not scroll vertically when the tooltips are added though, and looking around it seems like making the tooltips display outside of the parent div isn't possible with just CSS. 

Will try a few more things, otherwise may just have to go the modal route.

sei-aderr:
Couldn't get the css to cooperate in the scrolling div. Adjusted them all to modals

sei-aderr:
Apologies, didn't get the email that you had responded till after I adjusted them to be modals.

Either way, while the tooltips are a lot easier to mouse over, I think that the weird limitation with the div expanding looks really bad and just feels very unintuitive so I prefer the modal approach. I could adjust the modal to be the small version if you prefer that to what it looks like now.

Example here: https://patrickcate.github.io/vue-uswds/?path=/story/components-usamodal--default-modal

sei-aderr:
How it would look in this table,

image

PR #708, opened 2025-09-12
Feature/catalog changes

### LINKED ISSUE #699: Remove Code Type Field from Test Catalog Entries
We will no longer use the code type field, that used to have Measurement/Validation options. Now it will be assumed that all code in the test catalog is free form.

sei-secheverria:
Resolves #699 by removing the Code Type attribute from a Test Catalog Entry.

Also, adds frontend env cleaning to "make clean" task.

### DIFF
--- a/mlte/catalog/model.py
+++ b/mlte/catalog/model.py
@@ -1,6 +1,6 @@
 """Model implementation for MLTE catalog entries."""
 
-from typing import Any, Optional
+from typing import Optional
 
 from mlte.model import BaseModel
 from mlte.store.query import Filterable
@@ -57,7 +57,3 @@ class CatalogEntry(Filterable):
 
     def get_identifier(self) -> str:
         return self.header.identifier
-
-    def get_type(self) -> Any:
-        """Just to implement abstract class, not really needed."""
-        return "entry"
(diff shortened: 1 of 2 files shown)


mlte/catalog/model.py
```diff
@@ -78,4 +59,5 @@ def get_identifier(self) -> str:
         return self.header.identifier
 
     def get_type(self) -> Any:
```
sei-aderr, line 61: Is this needed at all if we don't have a type field anymore?

sei-aderr (dismissed):

mlte/catalog/model.py
sei-secheverria: A couple of classes, like this one and the Artifact one, derive from Filterable, a generic interface (actually abstract class) for things that can be filtered in searches/queries. "Type" was a common enough thing to include in that interface, as a shorthand. It still applies to Artifact, though it no longer applies to catalog entries. But since it is marked as an abstractmethod, it has to be implemented, even if it is empty. There are some ways around it: the base method could be not marked as abstractmethod, and we could just remove this from here; or we could remove the get_type method from Filterable and make the type be just one of the more generic get_property options, but this would involve more work. I'll chose the first option for clarity and simplicity.

Issue #590, opened 2025-01-27
Add CRUD for CustomLists


sei-secheverria:
Add a UI in the frontend, and endpoints in backend, to allow a user to modify the items in the CustomLists. For example, a user may want to add a new QA. One one hand, we want the user to be able to do this when, for example, adding something to the TestCatalog, in which case it may be added even only as a name. But if we want to add the description later, or modify name/description, or remove one, a CRUD will be needed.

Note that it is likely we won't keep this in the #505 epic, and this will be moved and be done later.

sei-secheverria:
Resolved in #721

Issue #362, opened 2024-02-26
Add the ability for user to add entries to drop downs when other is selected


sei-aderr:

sei-secheverria:
Resolved in #721

PR #721, opened 2025-09-25
Feature/custom list crud

sei-aderr:
Addresses #590, partially addresses #362

- Added CRUD for QAC/QA
- Added option to add QAC/QA within forms
- Made management of QAC/QA lists happen within state of application for easier sharing of data between components
- Updated backend to handle conversion of string url parameters to `CustomListName` 
- Fixed bug when redirecting upon saving a new negotiation card
- Re-organized pages in the frontend to have more structure


mlte/frontend/nuxt-app/components/CustomList/entry-list.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+  <table class="table usa-table usa-table--borderless">
+    <thead>
+      <tr>
+        <th scope="col" role="columnheader">Name</th>
+        <th scope="col" role="columnheader">Description</th>
+        <th scope="col" role="columnheader">Parent</th>
```
sei-secheverria, line 7: We may have to discuss this, but the way this list is showing ALL quality attributes, sorted by QA name, when the quality_attributes list is selected, is confusing. I think it would make more sense to group by parent; or at least, to sort by parent. This may be confusing if the parent is not the first column. We'd have to see how it looks, but maybe moving the Parent to the first column, and sorting first by parent and then alphabetically by child is a better way of showing information here. People looking at this list will usually not want to see all things, it makes more sense for people to want to see this items grouped by parent.

mlte/frontend/nuxt-app/components/CustomList/entry-list.vue
```diff
@@ -0,0 +1,38 @@
...
+        >
+          Delete
+        </UsaButton>
+      </td>
+    </tr>
+  </table>
+</template>
+
+<script setup lang="ts">
+const emit = defineEmits(["addEntry", "editEntry", "deleteEntry"]);
```
sei-secheverria, line 31: There is an issue here with changes and the initial list setup. Currently, the CUstomList store is set up in such a way that the "default" values for the lists (in this case QA cat and QAs) are added to the custom list each time a custom list store is connected/created/setup. That means that each time the backend is started, or each time a set_store() command is called, the CustomList store is filled with the initial default values, as seen in mlte/store/custom_list/initial_custom_list.py. This is a probelm when people edit values, but only in some cases. Adding a new entry to a list should not be an issue, as this setup only adds the default values, it does not clear what was already there. Losing edits to one of the initial entries should also not be an issue, as there is code in the initialization to just ignore the creation of a default list entry if it was already there. BUT the one case where there are issues is if one of the default list entries is deleted. In this case, this entry will be created again when the backend/set_store session is restarted, which is not expected behavior. We should discuss what to do to handle this case, as I don't have a clear answer.

mlte/frontend/nuxt-app/components/CustomList/entry-edit.vue
```diff
@@ -0,0 +1,128 @@
...
+    </UsaTextInput>
+
+    <UsaSelect
+      v-model="props.modelValue.parent"
+      :options="parentOptions"
+      :error="formErrors.parent"
+      :disabled="parentOptions.length == 0"
+    >
+      <template #label> Parent </template>
+      <template #error-message>Parent must be selected.</template>
```
sei-secheverria, line 34: Besides the issue previously discussed about the parent field not being updated when the add/edit page is loaded, the fact that the list that is the parent of this list is never shown. This field shows you the parent entry and allows you to change entries, but you can't see what is the name of the list these entries are coming from. Right now it is not an issue as there is only one parent list in existence, but later it could be confusing. I would add this information. One option would be to add another field, Parent List, that is not editable, where the parent list name is shown (or None if that list is mapped not to have any parent list).

mlte/frontend/nuxt-app/components/TestCatalog/entry-edit.vue
```diff
@@ -158,46 +158,32 @@ const formErrors = ref<Dictionary<boolean>>({
...
-  { name: "Sentiment Analysis", selected: false },
-  { name: "Regression", selected: false },
-  { name: "Segmentation", selected: false },
-  { name: "Tabular", selected: false },
-  { name: "Time Series", selected: false },
-]);
+const tagOptions = useTagOptions();
 
+await updateQAData();
+populateCatalogOptions();
```
sei-secheverria, line 164: There is a bug in how the tags are populated. If I edit (or maybe create as well) an entry, save, then add a new entry, it selects the tags that were selected for the previous entry, instead of starting from scratch.

mlte/frontend/nuxt-app/components/FormFields/system-requirements.vue
sei-secheverria, line 1: A comment both for the card and test entry: if a new item is added to a list, but the "Save" button is NOT clicked to save that new custom list item, and then I try to save the overall thing (card or test catalog entry), would it make sense to add a warning that prevents you from saving the overall artifact/entry and reminds you you have an unsaved custom list entry? It sounds like it might be a complex thing to do, but it seems to make sense from a user perspective.

sei-secheverria (changes requested):
Very nice, a couple of bugs, minor improvements, and some ideas.

mlte/backend/api/endpoints/custom_list_entry.py
```diff
@@ -136,11 +140,32 @@ def delete_custom_list_entry(
...
+
+
+@router.get("/{custom_list_id}/parent")
+def get_custom_list_parent(
+    *,
+    custom_list_id: str,
+    current_user: AuthorizedUser,
+) -> Optional[CustomListName]:
+    """
+    Get the name of parent custom list list.
```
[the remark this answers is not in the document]
sei-aderr: Adjusted

mlte/frontend/nuxt-app/components/CustomList/entry-edit.vue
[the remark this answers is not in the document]
sei-aderr: Fixed both issues.

Unfortunately I don't think there is a way to get rid of the `Select` option with the component we are using.

mlte/frontend/nuxt-app/components/CustomList/entry-list.vue
sei-aderr: This makes sense, adjusted it to be by parent, then by name within each parent group

mlte/frontend/nuxt-app/components/CustomList/entry-edit.vue
sei-aderr: Good catch, I added it into the label for the drop down so it will say `Parent: parentName`

mlte/frontend/nuxt-app/components/TestCatalog/entry-edit.vue
sei-aderr: Fixed

mlte/frontend/nuxt-app/components/FormFields/system-requirements.vue
sei-secheverria: If it sounds as it can be complex, I'd move it out to another ticket.

sei-aderr:
All comments are addressed, ready for another review.

sei-secheverria (approved):

PR #733, opened 2025-10-03
Task/non case sensitive search

### LINKED ISSUE #714: Make search in Test Catalog page not case sensitive
Currently the search for tags and QA are case sensitive and not very user friendly. Adjust the the backend to not discriminate on case

sei-aderr:
Resolves #714 

Add some frontend docstrings missed in last PR, make search non case sensitive.


mlte/store/query.py
```diff
@@ -153,11 +153,13 @@ class TagFilter(Filter):
...
-    value: Any
-    """The property to match."""
+    tag: str
+    """The tag to check against."""
 
     def match(self, item: Filterable) -> bool:
-        return self.value in item.get_tags(self.name)
+        return self.tag.lower() in [
+            tag.lower() for tag in item.get_tags(self.name)
+        ]
```
sei-secheverria, line 162: This is working pretty well, but I'd make an improvement. Currently, the property match works in a way that if the keyword being searched matches against any part of the property, it returns as a match. Which I think is great; if you search for "input", you get "Input", but also "Selected inputs and outputs", which is what I'd expect.

However, here, since tags is a list, the effect of ``self.tag.lower() in`` is that the exact tag (non case sensitive) is matched. So if I search for "vision" I don't get results for "Computer Vision", I have to write exactly "Computer Vision" (regardless of caps) to get that. This feels inconsistent between the two search boxes, and I also think it would be much more useful if the partial search works for tags as well.  Could you modify this to work in this way?

mlte/frontend/nuxt-app/pages/catalog/test-catalog.vue
sei-secheverria, line 1: This is working great, but I constantly miss a "Clear" button for when I quickly want to search something else, or for jumping between tags and qas and not searching for terms in both, without having to manually delete each.

Could we add a simple "Clear" button that simply clears both input fields?

sei-secheverria (changes requested):
It is working perfectly. However, I think there are two minor functional improvements we could add that greatly enhance the user experience, as described below.

sei-secheverria (approved):

Issue #734, opened 2025-10-03
Add support for multiple pieces of Evidence for a TestCase


sei-secheverria:
Currently TestCases can only have 1 piece of Evidence associated with them. This is restricted in several explicit ways:
 - Measurement's evaluate() and __call__() methods return only one Evidence.
 - Evidence's ids are assigned to be the id of the test case, not allowing multiple Evidences to be associated to one test case.

Modify how this works so that it is possible to generate one or more pieces of Evidences for any TestCase and with any Measurement.

sei-secheverria:
Ended up handling this in two ways:
 - There will not be any explicit support for returning more than one Evidence. By definition, each TestCase will have (and each Measurement will produce) one Evidence object, as it has a lot of metadata in it. If multiple values are needed, a new an Array evidence type can be used, or a new one can be implemented, to store more values.
 - ProcessMeasurements can now have an optional group id. This can be used to state that you want all measurements with he same group to be run at the same time on the same command, in parallel.
 - TestSuite.run_measurements will be modified so that it detects groups of ProcessMeasurements, and groups them first, and then runs each group as a whole. ProcessGroupMeasurement is now ProcessMeasurementGroup and will not be a Measurement, just a helper for manually running a group of measurements, or doing it through TestSuite.run_measurements

PR #735, opened 2025-10-03
Feature/process measurement improvements

### LINKED ISSUE #732: Add Better Method to run Multiple Measurements at the Same Time
Currently measurement classes are very isolated. It is assumed that each one will run separately. However, for measurements that gather data on resources being used by external processes that are running, it is likely you will want to run several such measurements at the same time. MLTE does not provide a quick way of doing this, though it does have a basic method for it. There are slightly different cases that could be handled, but this will focus on the following: Two or more measurements execute different tools to measure separate resources (eg, CPU and memory).

All measurements that use external tools derive from ProcessMeasurement. The only method currently available that allows you to avoid running the process you are measuring on multiple times is the evaluate_async method. In this case you start a process, call evaluate_async on each measurement, and then call wait_for_output on each measurement.

Example of how it currently looks like:

```
# Create measurements
cpu_measurement = LocalProcessCPUUtilization("predicting cpu")
mem_measurement = LocalProcessMemoryConsumption("predicting memory")

# Start the process to measure.
pid = ProcessMeasurement.start_script(MODEL_COMMAND[1], MODEL_COMMAND[2:])

# Execute the measurements
cpu_measurement.evaluate_async(pid)
mem_measurement.evaluate_async(pid)
cpu_stats: CPUStatistics = cpu_measurement.wait_for_output()
mem_stats: MemoryStatistics = mem_measurement.wait_for_output()
```

The recommendations to improve this are the following: It would be useful to have a way to group multiple measures and do the whole process of running them at the same time more automatic. Probably the simplest way to do this would be to abstract out the use of the start, eval_async and wait_for_output calls. A MeasurementGroup where you can add measurements, and then call evaluate, which will then start the process, evaluate and wait for input, making the process more transparent. It would look something like this from the outside:

```
# Create measurement group
measurements = ProcessMeasurementGroup()

# Add measurements to group.
measurements.add(LocalProcessCPUUtilization("predicting cpu"))
measurements.add(LocalProcessMemoryConsumption("predicting memory"))

# Evaluate the measurements.
command = [MODEL_COMMAND[1], MODEL_COMMAND[2:]]
measurements.evaluate(command)
```

And then internally that function would look like this:

```
class ProcessMeasurementGroup:
  measurements: list[ProcessMeasurement]

  def evaluate(self, command: list[]):
    # Start the process to measure.
    pid = ProcessMeasurement.start_script(MODEL_COMMAND[1], MODEL_COMMAND[2:])

   # Set up all measurements.
    for measurement in self.measurements:
         measurement.evaluate_async(pid)

    # Wait for results. This could be threaded, but not sure of the benefits.
    evidences: list[Evidence] = []
    for measurement in self.measurements:
         evidence = measurement.wait_for_output()
         evidences.append(evidence)
    
   return evidences
```

Some improvements that could be done on this idea: 1) This type of group measurement might be hard to handle in the case you want to automate measurement execution, as you would have one measurement linked to multiple cases. This could be handled better if we had one test case with multiple evidences associated to it, which we don't yet.

sei-secheverria:
Resolves #732 by creating a new class, ProcessGroupMeasurement, that hides the complexity of running multiple ProcessMeasurements that want to execute on the same external target process. 
 * Internal functionality was already available, but this simplifies setting things up by having one class and one evaluate() method that does the whole process of starting a process, getting the pid, and executing each measurement listed asynchronously. 
* This is not yet a real Measurement class as it returns multiple Evidences, something that is not yet supported by the framework, but will be worked on in the future (#734). This means that it can't be used to run automatic measurements with TestSuite.run_measurements(). However, it can be used manually, as shown in the demos, for easier setup.
* Also fixes minor demo issue setting up store multiple times.
* Also fixes Memory measurement crash when no captures are obtained for whatever reason.
* Adds capturing trace in error details when ProcessMeasurement tool throws exception.


sei-secheverria:
Ended up merging work for #734 here, as it was intertwined and more related to this PR.

sei-secheverria:
@sei-aderr sorry for the constant additions, PR should be ready to test now

mlte/measurement/process_measurement.py
```diff
@@ -8,49 +8,59 @@
...
 # -----------------------------------------------------------------------------
 # ProcessMeasurement
 # -----------------------------------------------------------------------------
 
 
-class ProcessMeasurement(Measurement, ABC):
+class ProcessMeasurement(Measurement):
     """Base class to be extended to measure external processes."""
 
+    PROCESS_GROUP_KEY = "group"
```
sei-aderr, line 28: Doesn't all caps for a variable usually indicate that it is a constant? Is there a reason you went with this?

sei-aderr: Nevermind, it is used as a constant and I misread the lines it was used in.

PR #754, opened 2025-10-13
Feature/rdbs custom list

### LINKED ISSUE #566: Implement Relational DB Custom List Store

sei-aderr:
Resolves #566, #763

Adds a RDBS store implementation for custom lists and improves variable naming in other RDBS stores.


mlte/store/custom_list/underlying/rdbs/metadata.py
```diff
@@ -0,0 +1,29 @@
...
+class DBBase(DeclarativeBase):
+    """Base class for DB table classes."""
+
+    pass
+
+
+class DBCustomListEntry(DBBase):
+    __tablename__ = "custom_list_entry"
+
+    list_name: Mapped[CustomListName]
```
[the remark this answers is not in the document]
sei-secheverria: To be followed up in #764

mlte/store/custom_list/underlying/rdbs/store.py
```diff
@@ -84,18 +84,28 @@ def create(
...
+                if new_entry.parent:
+                    _, parent_orm = DBReader.get_entry_by_name(
+                        new_entry.parent, session
+                    )
+
+                entry_orm = create_custom_list_entry_orm(
+                    new_entry,
+                    list_name,
+                    parent_id=parent_orm.id if parent_orm else None,
+                )
```
sei-secheverria: There is a way to simplify this, to avoid having to get the parent manually each time, and get SQLAlchemy to get the id for you. I'll elaborate on the DB metadata module.

mlte/store/custom_list/underlying/rdbs/reader.py
```diff
@@ -30,30 +28,69 @@ def get_entry(
...
+            )
+        else:
+            parent_orm = None
+            if entry_orm.parent:
+                _, parent_orm = DBReader.get_entry_by_id(
+                    entry_orm.parent, session
+                )
+
+            return (
+                create_custom_list_entry_model(
```
sei-secheverria: As well as this one

mlte/store/custom_list/underlying/rdbs/metadata.py
```diff
@@ -17,13 +17,14 @@ class DBBase(DeclarativeBase):
...
+    id: Mapped[int] = mapped_column(primary_key=True)
     list_name: Mapped[CustomListName]
-    name: Mapped[str] = mapped_column(primary_key=True)
+    name: Mapped[str]
     description: Mapped[str]
 
-    parent: Mapped[Optional[str]] = mapped_column(
-        ForeignKey("custom_list_entry.name"), nullable=True
+    parent: Mapped[Optional[int]] = mapped_column(
+        ForeignKey("custom_list_entry.id"), nullable=True
```
sei-secheverria: Ok, so this is fine and it is the way it should be. However, if you see for example how it is implemented in, let's say, DBArtifact, for the model, you have something like this:

```
  model_id: Mapped[Optional[int]] = mapped_column(ForeignKey(DBModel.get_id_column())  )
  model: Mapped[Optional[DBModel]] = relationship()
```

We have an explicit id foreign key, as well as a "fake" column, that gets populated from that foreign key. In our case, I'd rename this to parent_id, and then add a new ``parent: Mapped[Optional[DBCustomListEntry]] = relationship()``. 

Now, when this is setup, the other table, in the example above DBModel, has a corresponding relationship field that backpopulates things like the id above. In the case of DBModel, you can see this:

```
    artifacts: Mapped[list[DBArtifact]] = relationship(back_populates="model", cascade="all, delete-orphan")
```

Where "model" points to the relationship field in DBArtifact that is linked to this. This, in our case, we would also have to add, to CustomListEntry, the following field (or something like this):

```
    children: Mapped[list[DBCustomListEntry]] = relationship(back_populates="parent", cascade="all, delete-orphan")
```

Then, later, you can forget about the parent_id column, except inside select queries to do joins, if applicable, and just use the parent one as a model. Both columns will cross-populate if it is properly set up. We can discuss this in more detail if it is too complex. But once you have this set up, there ends up being a more transparent relationship between parent and children, even with this made-up child list of entries, and it avoids having to deal with the id column every time.

sei-aderr: Thank you for this explanation. I was working on reducing that duplication and it wasn't being simple, so it makes sense I was missing something. Adding the "fake" column as well as the child makes a lot of sense and I added those in. This let me delete the grab by id and parent id manipulation I was doing in the `reader.py`. 

I am not sure how to make the change in `store.py` work though and keep the same file structure. When getting to `create` in `store.py` all that method receives is the name of the parent, so it needs to grab the parent orm to create the entry orm. So it looks like in `main_factory`, it just uses the `DBReader` to get data for the orm that it is making, but I don't see how it is avoiding a circular import because the factory imports from the reader and the reader imports from the factory. When I set mine up that way, I get a circular import error. It looks like the catalog store avoids this by having the factory within the reader file. Do you know how the artifact store implementation is avoiding that?

sei-secheverria (approved):

PR #773, opened 2025-10-21
Feature/add nbconvert

### LINKED ISSUE #746: Design method to create sample catalog entries from demos
The catalog entries, as we define them now, are mostly going to be composed of the demo Jupyter notebooks that are associated with the different QAs. However, the current process of creating those sample catalog entries is manual an painful. Think about the following:
 - Is this exactly what we need in the TestCatalog? Is just putting each notebook there enough? Are we missing something?
 - Is there a way to create a script that can automatically update the sample catalog entries out of the demos? How would that look like, and what would it require?
 - In a related but separate issue, how should we show this in the UI? Currently, the code we show in the UI is very ugly, as it is just a text area. If we are going to be showing the notebook demo code, is there at least some way we can add colored syntax highlights to the UI so it is easier to read?

sei-aderr:
Resolves #746 

Locks demo dependency versions, adds docs about nbconvert to be used to convert demo notebooks into test catalog entries


poetry.lock
sei-secheverria: Do you know why all these things were updated? We didn't change any dependencies, other than removing the caret.

docs/docs/development.md
```diff
@@ -63,6 +63,12 @@ $ cd demo
 $ bash run_environment.sh
 ```
 
+When demo notebooks have been created and need to be added to the test catalog, they can be converted to a script using `nbconvert`. This tool is available is available if the demo dependencies are installed.
+
+```bash
+$ jupyter nbconvert --to script demo/simple/1_negotiation.ipynb
```
sei-secheverria: It would be great to have a make command here, but make commands are fiddly about arguments, so I get why this is explicit here for now.

I would change the sample notebook. We will only want to create entries from the evidence notebooks, so please use one of the 2a_... or similar notebooks as an example, from the scenarios demo.

PR #775, opened 2025-10-23
Feature/automated nbconvert

### LINKED ISSUE #772: Implement automated conversion of notebook to catalog entries
Continuation of #746 

Plan is to use `nbconvert` to go from notebook to catalog entry code block. While this will work, it will currently just be a manual usage of the tool to convert them. 

It would be nice to automate this process to do all the conversions at once, but will require a couple tings
- Write script to run nbconvert on each of the demo notebooks
- Add call to script in makefile
- Convert each of these scripts into the single line format to insert into the test catalog entry json
- Map each notebook to a test catalog entry so that they can all be ran at once

Then we need to consider adding this to qa/ci to ensure all the entries are always updated, but need to consider overhead and the frequency of demo updates

sei-aderr:
Resolves #772 

- Adds script that goes through the `scenarios` demo and updates all data about the test catalog entry it is related to.
- Adds needed data to the top section of each `scenarios` evidence demo
- Script also has a check mode to ensure that all test catalog entries are updated with the CI

### DIFF
--- a/demo/catalog_entries.py
+++ b/demo/catalog_entries.py
@@ -14,11 +14,14 @@ import sys
 import tempfile
 import time
 from pathlib import Path
+from typing import Any
 
 import nbformat
 from nbformat import NotebookNode
 
 from mlte.catalog.model import CatalogEntry, CatalogEntryHeader
+
+SAMPLE_CATALOG_PATH = "../mlte/store/catalog/sample"
 
 
 def main():
@@ -34,13 +37,12 @@ def main():
     timestamp = int(time.time())
     notebook_path = Path(sys.argv[2])
     demo_name = notebook_path.parent
-    tmp_script_path = Path(f"/tmp/mlte/{demo_name}/{notebook_path.stem}.py")
-    entry_qac = tmp_script_path.name.split("evidence_")[-1][:-3]
-    entry_identifier = f"{demo_name}-{entry_qac}"
-    entry_path = Path(f"../mlte/store/catalog/sample/{entry_identifier}.json")
+    entry_qa = notebook_path.name.split("evidence_")[-1][:-6]
+    entry_identifier = f"{demo_name}-{entry_qa}"
+    entry_path = Path(f"{SAMPLE_CATALOG_PATH}/{entry_identifier}.json")
 
     notebook_data, notebook_entry_json = read_notebook(notebook_path)
-    code_str = create_code_str(notebook_data, tmp_script_path)
+    code_str = create_code_str(notebook_data)
     new_entry = create_entry(
         notebook_entry_json, entry_identifier, timestamp, code_str
     )
@@ -52,7 +54,7 @@ def main():
     if mode == "check":
         if not current_entry or not compare_entries(new_entry, current_entry):
             print(
-                f"Sample Catalog Entry: {entry_qac}, in demo {demo_name} is not updated."
+                f"Sample Catalog Entry: {entry_qa}, in demo {demo_name} is not updated."
             )
             sys.exit(1)
         else:
@@ -108,7 +110,7 @@ def read_notebook(notebook_path: Path) -> tuple[NotebookNode, dict]:
 
 
 def create_entry(
-    notebook_entry_json: dict,
+    notebook_entry_json: dict[str, Any],
     entry_identifier: str,
     timestamp: int,
     code_str: str,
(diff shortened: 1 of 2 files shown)


demo/catalog_entries.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+# Build mode: Converts demo evidence notebooks into scripts, and adds or updates them in the sample test catalog
+# Check mode: Checks that all sample test catalog entries are updated
+
+if [ "$1" != "build" ] && [ "$1" != "check" ]; then
+    echo "Invalid mode."
+    exit 1
+fi
+
+for notebook_file in ./scenarios/*ipynb
```
sei-secheverria: This is assuming this will only be done for the scenarios demo, and will not easily scale to the new demos we want to apply this on. Similar to what we have in the test.sh script, you could simply have a list of the demo folders that this apply to, and and add a top level loop to iterate over them, and do all of this to each of the demos we want to apply this to.

demo/catalog_entries.py
```diff
@@ -0,0 +1,151 @@
+# Build mode: Take demo notebook script conversions and put them into the sample catalog
+# Check mode: Checks that all sample test catalog entries are updated
```
sei-secheverria: I'd change this format to use the standard docstring for modules. You can add a short explanation, followed by these modes.

demo/catalog_entries.py
```diff
@@ -0,0 +1,151 @@
...
+import tempfile
+import time
+from pathlib import Path
+
+import nbformat
+
+from mlte.catalog.model import CatalogEntry, CatalogEntryHeader
+
+
+def compare_entries(entry1: CatalogEntry, entry2: CatalogEntry) -> bool:
```
sei-secheverria: I'd add docstring here too

demo/catalog_entries.py
```diff
@@ -0,0 +1,151 @@
...
+    if (
+        entry1.header.creator == entry2.header.creator
+        and entry1.header.updater == entry2.header.updater
+        and entry1.header.catalog_id == entry2.header.catalog_id
+        and entry1.tags == entry2.tags
+        and entry1.quality_attribute == entry2.quality_attribute
+        and entry1.code == entry2.code
+        and entry1.description == entry2.description
+        and entry1.inputs == entry2.inputs
+        and entry1.output == entry2.output
```
sei-secheverria: Is there a reason we have to do this part by part? Is it to explicitly ignore something, like timestamps?

demo/catalog_entries.py
```diff
@@ -0,0 +1,151 @@
...
+    ):
+        return True
+    else:
+        return False
+
+
+if __name__ == "__main__":
+    if len(sys.argv) < 3:
```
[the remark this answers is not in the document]
sei-secheverria: Also, this main function looks a bit too long and hard to follow.. I'd try to split it into some subfunctions for clarity.

demo/catalog_entries.py
```diff
@@ -0,0 +1,151 @@
...
+    entry_file_name = SCRIPT_PATH.name.split("evidence_")[-1][:-3]
+    entry_path = Path(
+        f"../mlte/store/catalog/sample/demo-{entry_file_name}.json"
+    )
+    if os.path.exists(entry_path):
+        with open(entry_path, "r") as entry_file:
+            current_entry = CatalogEntry.from_json(json.load(entry_file))
+
+    if MODE == "check":
+        if not compare_entries(new_entry, current_entry):
```
sei-secheverria: current_entry only exists as a variable if the entry_path path already existed. What will happen to this code if not? current_entry should likely be initialized to None or similar, and checked against that where needed.

demo/catalog_entries.py
```diff
@@ -0,0 +1,151 @@
...
+            print(f"Sample Catalog Entry: {entry_file_name}, is not updated.")
+            sys.exit(1)
+        else:
+            sys.exit(0)
+
+    elif MODE == "build":
+        if not compare_entries(new_entry, current_entry):
+            with open(entry_path, "w") as entry_file:
+                json.dump(new_entry.to_json(), entry_file, indent=4)
```
sei-secheverria: Will this only update existing entries? What happens for new demos or new entries? I think we should define that we will create entries for all 2_ notebooks, regardless on whether they already existed or not.

demo/catalog_entries.sh
```diff
@@ -1,23 +1,36 @@
...
-        elif [[ "$1" == "check" ]]; then
-            python catalog_entries.py check $notebook_file
-            if [[ $? == 1 ]]; then
-                exit $?
+    for notebook_file in ${demo}/*ipynb
+    do
+        # Get the file name without the demo path prepended
+        NAME=${notebook_file##*$demo/}
+        # Check if the first character of the path is a 2, and the second is a lowercase letter
+        #   This is the naming format for QAC demos that will be enforced so we can identify them here
```
sei-secheverria: I'd change this comment from "QAC" to "QA" for clarity.

Issue #770, opened 2025-10-17
Add demo-test to CI


sei-secheverria:
Add demo-test command to CI. Check strange json_fix error that shows up when importing matplotlib, which only happens in Github Actions and not locally.

Sample failed run: https://github.com/mlte-team/mlte/actions/runs/18689552085/job/53291338484

PR #776, opened 2025-10-23
Dependency bug fixes and demo-test integration into Github Actions

### LINKED ISSUE #774: Bug installing psycopg-binary in MacOS with Python 3.9
When using Python 3.9 and MacOS, psycopg-binary library fails with unclear error. This is a known error on that library (https://github.com/psycopg/psycopg2/issues/1737) . Apparently this was only on version 2.9.10, but it was fixed in 2.9.11. Update version to fix this.

sei-secheverria:
Fixes #774 and #770. More specifically:
 - For #774, psycopg-binary version was updated to 2.9.11, to avoid bug in previous version that caused issue in Linux and Python 3.9.
 - For #770, added demo-test command to CI action. To allow this to work, some Linux-specific fixes were needed:
   - Updated json-fix version to 1.0.2 (it was just published after we requested it), which fixes incompatibility with matplotlib and json_fix due to explicit JSONencoder, that was only triggering in Linux. 
   - Fixed sorting order of notebook files to ensure they are processed in the proper order in Linux as well as in MacOS.
   - Changed slightly copy command to seed store demo from sample_store to properly copy things over in Linux; also separated the command used in test from the one used in the notebook itself to avoid a weird relative folder issue.


demo/setup_store.sh
sei-aderr: Why was this removed, and then the code was duplicated the two separate places? I see it mentioned in the PR, they look the same but were they acting differently when using the script?

Issue #758, opened 2025-10-14
Add documentation about overall demo structures


sei-secheverria:
For the demo, we are following a structrure of having one or two notebooks setting up a card and suite, several more as evidence, etc. However, this is not written down anywhere, so there is no guidance on how to create a new demo, there are some things that are slightly different between each other, and no clear guidance on how to walk through the demos either (this last point more closely related to #757 ). Update documentation to include this.

PR #785, opened 2025-10-31
Feature/demo docs

### LINKED ISSUE #757: Update docs to better guide people starting development
Currently, the readme is slightly more geared towards people downloading the MLTE package, rather than people developing. Though there is a development.md doc that is referenced there that has a breath of information about development, getting started with development can be unclear or daunting. Improve documentation to 1) make it clear how people working on the repo should work, maybe a better summary in the readme of what is in development.md, and 2) putting the demos up front in more detail as THE way to start getting into MLTE and testing things out, including how to play with the notebooks and running the run_environment script for the backend and frontend.

sei-aderr:
Resolves #757, #758

Clear up dev and demo docs.

### DIFF
--- a/docs/docs/development.md
+++ b/docs/docs/development.md
@@ -101,7 +101,7 @@ $ source .venv/bin/activate
 
 There are a couple of shorthand commands in the Makefile to run several of the below commands at the same time. The most useful ones include:
 
-* `make qa`: executes the schema generation, doc check, source sorting, formatting, linting, static type checking commands of Python and Typescript code, and sample test catalog generatiion.
+* `make qa`: executes the schema generation, doc check, source sorting, formatting, linting, static type checking commands of Python and Typescript code, and sample test catalog generation.
 * `make check-qa`: executes the schema check, doc check, source sorting check, formatting check, linting check, static type checking commands of Python and Typescript code and sample catalog check.
 * `make ci`: executes the same commands as `check-qa`, but also runs `test` to execute the unit tests, cleaning caches first to better simulate execution in a CI environment.
 
--- a/docs/docs/new_demo.md
+++ b/docs/docs/new_demo.md
@@ -15,20 +15,20 @@ A demo is a set of Jupyter notebooks that walk through the MLTE IMT process. Dem
     - `<x>` being letters starting at `a` and continually incrementing, for example `2a_evidence_fairness`, `2b_evidence_robustness`, `2c_evidence_resilience`.
     - Each notebook should gather evidence for all the `TestCase`s that test a quality attribute scenario from the [Negotiation Card](negotiation_card.md) that relate to the quality attribute in the name of the notebook.
     - Each of these notebooks will be also used as an entry in the Sample Test Catalog to give an example of the `Test Case`s evaluated in the notebook. The second cell must contain a JSON block with information that will be used to populate the Test Catalog entry. All fields are required.
-        - Tags: System-defined tags that are used in catalog search to group the catalog entries. Current options are found in `mlte/frontend/nuxt-app/composables/state.ts`as the const `useTagOptions`. These will be updated to be a custom list in the near future at which point a list of options will be found in `mlte/store/custom_list/tags`.
-        - Quality Attribute: High-level quality attribute category that the notebook is validating. Options can be found in `mlte/store/custom_list/quality_attributes` and should match the quality attribute in the notebook name.
-        - Description: Description of the scenario that this notebook is testing.
-        - Inputs: Inputs to the model that are being tested.
-        - Outputs: Expected output from the test
-        ```json
-        {
-            "tags": ["General"],
-            "quality_attribute": "Detect OOD inputs and shifts in output",
-            "description": "During normal operation, the ML pipeline will log errors when out of distribution data is observed. The ML pipeline will create a log entry with a tag. During normal operation, ML pipeline will log errors when the output distribution changes. The ML pipeline will create a log entry with a tag.",
-            "inputs": "Existing ML model, sample image data that has out of bounds input, and that produces output confidence error",
-            "output": "Log with input issues tagged",
-        }
-        ```
+        - `Tags`: System-defined tags that are used in catalog search to group the catalog entries. Current options are found in `mlte/frontend/nuxt-app/composables/state.ts`as the const `useTagOptions`. These will be updated to be a custom list in the near future at which point a list of options will be found in `mlte/store/custom_list/tags`.
+        - `Quality Attribute`: High-level quality attribute category that the notebook is validating. Options can be found in `mlte/store/custom_list/quality_attributes` and should match the quality attribute in the notebook name.
+        - `Description`: Description of the scenario that this notebook is testing.
+        - `Inputs`: Inputs to the model that are being tested.
+        - `Outputs`: Expected output from the test.
+            ```json
+            {
+                "tags": ["General"],
+                "quality_attribute": "Detect OOD inputs and shifts in output",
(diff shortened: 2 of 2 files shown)


docs/docs/development.md
```diff
@@ -2,6 +2,26 @@
...
+
+The best examples of how to use MLTE are contained with in the [Demos](#demos). If looking to get started on development, these are the best place to start to get a feel for how the tool works. The code within the demos walks through the MLTE library IMT process. These can be ran after making a virtual environment and installing `MLTE` along with the demo dependencies.
+
+```bash
+$ pyenv install 3.9
+$ pyenv local 3.9
+$ make venv
+```
+
+The other part of MLTE is the frontend and backend. These are used for the SDMT process, and visualing results from the IMT process. This can be setup as explained in the demo section by using the `demo/run_environment.sh` script. This will start `MLTE` as 3 docker containers with an RDBS based backend. The frontend will be available at `localhost:8000` and will include a sample model, version, and negotiation card.
```
sei-secheverria: Again, there is some confusion here between IMT and SDMT. The frontend/backend are used in the SDMT process, for visualizing the results, but again, the results of the SDMT process. The only result of IMT is the Negotiation Card, and though, yes, you can see it in the frontend/backend, at most you would use them near the end of IMT, to create/update the card, and then on for the SDMT process. You could say "These are used for the SDMT process, and visualizing results from the this process, as well as providing a user-friendly interface to create the Negotiation Card in the first place" (or something along those lines.

docs/docs/development.md
```diff
@@ -173,69 +234,60 @@ $ make check-schema
...
+We format and lint all .vue, .js, and .ts files with <a href="https://eslint.org/" target="_blank">ESLint</a>, which can be run from the root of the repository with:
 
 ```bash
-$ npm install
-$ npx gulp init
+$ make lint-frontend
 ```
 
-You can also run the following make command:
+Or manually from the root of the nuxt application:
```
sei-secheverria: I would remove all of these "or manually" parts from the frontend. We used to have them for the python code, but it was a pain to maintain, and also it is duplicated; if someone wants to see how to manually do this, they can check in the Makefile. And honestly, I'd prefer people to just run the make commands in general to avoid errors; if they really can't or need to do something different, they will have enough knowledge to get the details from the makefile.

docs/docs/development.md
```diff
@@ -63,7 +83,38 @@ $ cd demo
...
+
+A demo is a set of Jupyter notebooks that walk through the MLTE IMT process. Demos are to be named and populated as follows. The number naming is used to signify the order in which the notebooks should be ran.
+
+0. `all notebooks`
+    - All notebooks will have to connect setup the `MLTE` context before their operation. This is done by importing the session module, `from demo.scenarios.session import *`
+    - All files outside of notebooks including data, or re-used functionality should be included within the demo folder
+1. `1_requirements.ipynb`
+    - Defines the  `MLTE` [Negotiation Card](negotiation_card.md), an example card can be found at `./demo/sample_store/models/OxfordFlower/card.default.json`
+    - Defines the `TestSuite`, more `TestSuite` information can be found in [Using `MLTE`](using_mlte.md#testing-models-with-`mlte`-(imt-and-sdmt))
+2. `2a_evidence_<quality_attribute> - 2<x>_evidence_<quality_atribute>`
```
sei-secheverria: There seems to be some duplication here. It says "`2a_evidence_<quality_attribute> - 2<x>_evidence_<quality_atribute>`", but shouldn't it be just "`2<x>_evidence_<quality_atribute>`"?

docs/docs/development.md
```diff
@@ -63,7 +83,38 @@ $ cd demo
...
+    - Naming of match the scheme, for example `2a_evidence_fairness`, `2b_evidence_robustness`, `2c_evidence_performance`.
```
sei-secheverria: "Naming should match the scheme" maybe?

docs/docs/development.md
```diff
@@ -63,7 +83,38 @@ $ cd demo
...
+            "tags": ["General"],
+            "quality_attribute": "Detect OOD inputs and shifts in output",
+            "description": "During normal operation, the ML pipeline will log errors when out of distribution data is observed. The ML pipeline will create a log entry with a tag. During normal operation, ML pipeline will log errors when the output distribution changes. The ML pipeline will create a log entry with a tag.",
+            "inputs": "Existing ML model, sample image data that has out of bounds input, and that produces output confidence error",
+            "output": "Log with input issues tagged",
+        }
+        ```
+3. `3_report.ipynb`
+    - Create a `TestSuiteValidator` and validate the evidence collected.
+    - Generate a report to communicate the results of the evaluation.
```
sei-secheverria: This is not really the structure we are currently following for the new demos, where we have several sub 3 reports (for validation, manual validation and report generation). Make sure this is in line with the new demos.

docs/docs/development.md
```diff
@@ -63,7 +83,38 @@ $ cd demo
...
+
+All other files related to the new demo including data
```
sei-secheverria: This seems to be a leftover that was moved higher up

docs/docs/development.md
```diff
@@ -63,7 +83,38 @@ $ cd demo
 $ bash run_environment.sh
 ```
 
-When demo notebooks have been created and need to be added to the test catalog, updated with the following command. This requires the demo dependencies to be installed.
+#### Creating a new Demo
```
sei-secheverria: If this section ends up being too large, it is ok to separate it to another .md document. It may be much cleaner that way.

sei-secheverria (changes requested):
Very good, overall nice work! There are however some errors and some missing info to add to the Demo section.

docs/docs/development.md
sei-secheverria: Oh, yeah, that makes sense. But yeah, I'd change it, and explain that "where 'x' is a letter from a to whatever"

docs/docs/development.md
```diff
@@ -136,8 +101,8 @@ $ source .venv/bin/activate
 
 There are a couple of shorthand commands in the Makefile to run several of the below commands at the same time. The most useful ones include:
 
-* `make qa`: executes the schema generation, doc check, source sorting, formatting, linting, and static type checking commands of frontend and backend.
-* `make check-qa`: executes the schema check, doc check, source sorting check, formatting check, linting check, and static type checking commands of frontend and backend.
+* `make qa`: executes the schema generation, doc check, source sorting, formatting, linting, static type checking commands of Python and Typescript code, and sample test catalog generatiion.
```
sei-secheverria: Small typo: "generatiion"

docs/docs/new_demo.md
```diff
@@ -0,0 +1,47 @@
...
+    - Each of these notebooks will be also used as an entry in the Sample Test Catalog to give an example of the `Test Case`s evaluated in the notebook. The second cell must contain a JSON block with information that will be used to populate the Test Catalog entry. All fields are required.
+        - Tags: System-defined tags that are used in catalog search to group the catalog entries. Current options are found in `mlte/frontend/nuxt-app/composables/state.ts`as the const `useTagOptions`. These will be updated to be a custom list in the near future at which point a list of options will be found in `mlte/store/custom_list/tags`.
+        - Quality Attribute: High-level quality attribute category that the notebook is validating. Options can be found in `mlte/store/custom_list/quality_attributes` and should match the quality attribute in the notebook name.
+        - Description: Description of the scenario that this notebook is testing.
+        - Inputs: Inputs to the model that are being tested.
+        - Outputs: Expected output from the test
```
sei-secheverria: Missing end period

sei-secheverria (approved):

PR #806, opened 2025-11-12
Refactor/parallel ci

### LINKED ISSUE #778: Separate Github CI flows
Current github CI flow job does too much; demo and frontend can be done separately and in parallel. Extract at least these two into parallel jobs.

sei-aderr:
Resolves #778 

Separated frontend and demo jobs out from the main. Added a first setup job that sets the env variables as json values because you can't set anything except strings in the env vars. It is one of the official way to handle this situation from the github docs. Went from 17-20min runs down to about 10mins so pretty solid improvement

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -18,10 +18,10 @@ jobs:
     steps:
       - name: Define os matrix
         id: os_matrix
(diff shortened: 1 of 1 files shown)


sei-secheverria (approved):
Very nice!

PR #813, opened 2025-11-18
Feature/simple demo extension

### LINKED ISSUE #804: Create Separate Demo based on Simple one for Interpretability
Create a new, very simple demo, similar in style and structure to GardenBuddy, but focusing on only one QA, Intepretability. Use the Iris model from SimpleDemo, as well as the negotiation card from it as a starting point. Do note that the NegotiationCard for the simple demo is not currently in the sample_store, as it is created in the notebook itself, so you probably want to run the simple demo, get the JSON negotiation card from the demo/store, and then modify it as needed and add it to the demo sample_store. Note that the model name for this demo should be different than the model name from the simple demo, even if both are using Iris underneath.

sei-browersinning:
Added interpretability to simple demo, with a manually generated test catalog entry.
I also fixed the tagging on the analyzability and monitorability notebooks. 

Resolves #804

### DIFF
--- a/demo/simple/0_negotiation.ipynb
+++ b/demo/simple/0_negotiation.ipynb
@@ -100,7 +100,7 @@
     "    data=[\n",
     "        DataDescriptor(\n",
     "            description=\"Iris flower dataset.\",\n",
-    "            purpos=\"Training\",\n",
+    "            purpose=\"Training\",\n",
     "            classification=DataClassification.UNCLASSIFIED,\n",
     "            access=\"None\",\n",
     "            labeling_method=\"by hand\",\n",


demo/GardenBuddy/2i_evidence_analyzability.ipynb
```diff
@@ -19,11 +19,11 @@
    "outputs": [],
    "source": [
     "{\n",
-    "    \"tags\": [],\n",
-    "    \"quality_attribute\": \"\",\n",
-    "    \"description\": \"\",\n",
-    "    \"inputs\": \"\",\n",
-    "    \"output\": \"\",\n",
+    "    \"tags\": [\"Computer Vision\", \"Object detection\"],\n",
+    "    \"quality_attribute\": \"analyzability\",\n",
```
sei-aderr: This needs to be capitalized `Analyzability` to work with properly. Also capitalize the start of description, inputs, output

demo/GardenBuddy/2j_evidence_monitorability.ipynb
```diff
@@ -18,10 +18,10 @@
    "source": [
     "{\n",
     "    \"tags\": [\"Computer Vision\", \"Object detection\"],\n",
-    "    \"quality_attribute\": \"analyzability\",\n",
-    "    \"description\": \"\",\n",
-    "    \"inputs\": \"\",\n",
-    "    \"output\": \"\",\n",
+    "    \"quality_attribute\": \"monitorability\",\n",
```
sei-aderr: This needs to be capitalized `Monitorability` to work with properly. Also capitalize the start of description, inputs, output

demo/simple/0_negotiation.ipynb
```diff
@@ -34,16 +37,26 @@
...
     "\n",
     "In MLTE, we negotiation requirements with the help of a `NegotiationCard`."
    ]
   },
+  {
+   "cell_type": "markdown",
+   "id": "3708f9de",
+   "metadata": {},
+   "source": [
+    "The context for this system is that Iris flower experts are working as stakeholders with a model development team to build a trained ML model that will correctly classify Iris flowers from physical measurements. Botanists use a dichotomous key for Iris classification."
```
sei-secheverria: I moved some of the context from one of the scenarios up here for clarity. Would it be possible to add a reference to the model or model inspiration here, as we've done for the other demos?

PR #816, opened 2025-11-21
Feature/inter store validation

### LINKED ISSUE #782: Add Support for Inter-Store Validation
As described in #728, we want to be able to validate that strings that come from CustomLists are actually coming from CustomLists. The reason this validation is not automatic is due to the stores being completely independent. The three main current cases where there is a connection is between the Artifact Store header user, and a user's username in the User Store; and between the CustomList entries in the CustomList Store, and values in the Negotiation Card in the Artifact Store, and in a Catalog Entry in the Catalog Store. The independence of the stores prevents easy cross validation of values that end up a strings, but for several reasons, unifying all stores is not currently feasible.

A potential way to add support for inter-store validation would be the following:
 - Leverage the SessionStores class, which holds all the stores being used either by the running backend on a library session, and add methods to perform the validation there, since that class already knows and has access to all stores.
 - To still allow the stores to be used directly, a generic validation hook could be added to all CRUD methods in the stores, and that hook could be setup/injected by SessionStores when setting up the stores, injecting access to validation methods it has, so they are called transparently when needed by the stores. This could be done in several ways. One way is as follows:
    - Create a CrossValidator "interface" that has a .validate_<action>() method for each CRUD action, and receives stores in its constructor.
    - CrossValidators should be composable, so that a series of validations can be performed by just one call, possible through a CompositeValidator class or similar.
    - Add a CompositeValidator to the ResourceMapper class, with no validators by default.
    - For each store, ensure that each CRUD method of each ResourceMapper calls the corresponding .validate_<action>() method of its CrossValidator when it starts. This could be done calling it in the base ResourceMapper class, and ensuring each class calls the super() method for each CRUD action.
     - Create CrossValidators for the three existing cases: validate an Artifact against the User store, validate an Artifact against the CustomList store, and validate a CatalogEntry against the CustomList store.
     - In SessionStores, add those validators to the CompositeValidator of the required store's Resource Mappers (in this case, Artifact mapper, and CatalogEntry mapper).

Note that for this to work transparently in the ArtifactStore, #581 should be addressed first.

### LINKED ISSUE #797: Add inter-store validation of CustomList entries
Using what is being done in #782, add inter-store validation of CustomList entries. Right now, it would be needed for the NegotiationCard and the CatalogEntry

sei-aderr:
Resolves #782, resolves #797

Implementation of inter store validation for users and custom list entries in both artifacts and catalogs. largely finished, but some things to still discuss/work on

- [x] Couple TODOs added to discuss, mainly about accepting default values for fields in validators
- [x] Finish full unit test implementation for the validators. Currently have it working for memory

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -63,6 +63,8 @@ jobs:
         python-version: ${{ fromJSON(needs.define-matrix.outputs.py_ver_matrix) }}
     runs-on: ${{ matrix.os }}
     steps:
+      - name: Clear space from tools folder, not needed because this runs in docker
+        run: rm -rf /opt/hostedtoolcache
       - uses: actions/checkout@v4
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v5


mlte/store/validators/cross_validators.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def validate(self, new_artifact: ArtifactModel) -> None:
+        if self.user_store is None:
+            raise RuntimeError(
+                "Artifact user validator's user store has not been set."
+            )
+
+        if new_artifact.header.creator:
+            with ManagedUserSession(self.user_store.session()) as session:
+                # TODO: Make sure this is always valid, it is what we do in the simple demo, maybe some other places
```
sei-secheverria: Does the TODO refer to line 37? If so, it is fine. The default value for that model field is to be None anyway. Not sure what the question here is.

mlte/store/validators/composite_validator.py
```diff
@@ -0,0 +1,55 @@
...
+        """Validate all validators."""
+        if self.validators:
+            for validator in self.validators:
+                validator.validate(new_resource)
+
+
+class CrossValidator(ABC):
+    """Interface to define a CrossValidator that validates store entries against data in separate stores."""
+
+    # TODO: These are Any to avoid circular imports, can this be added somehow?
```
sei-secheverria: I think the circular import would be fixed simply by separating the composite validator from the cross validator into two different modules. The stores never import the crossvalidator, they just import the composite one. It just happens that both are in the same module.

mlte/session/session_stores.py
```diff
@@ -111,8 +119,20 @@ def setup_stores(
     custom_list_store = InitialCustomLists.setup_custom_list_store(stores_uri)
     stores.set_custom_list_store(custom_list_store)
 
+    # Setup catalog store validators
```
sei-secheverria: This whole function is ending up too large, with too many sub parts. I'd split it into subfunctions to set up each type of store (setup_artifact_store, setup_catalog_stores, etc), and have them all be called from here. It would group together the setting up of the stores and the validators. And in the case of the catalogs, the extra weird stuff it does.

mlte/store/validators/composite_validator.py
sei-aderr: The circular import is caused by `store/artifact/store.py` if `from mlte.store.artifact.store` is imported. Because `composite_validator > artifact_store > base > composite_validator`. Same with the other store types, would moving it out to a separate module fix that in some way?

mlte/session/session_stores.py
```diff
@@ -132,4 +152,26 @@ def setup_stores(
...
+    stores.artifact_store.set_validators(artifact_store_validators)
+
+    # Add catalog store validators to stores
+    for uri, catalog_store in stores.catalog_stores.catalogs.items():
+        if uri == SessionStores.LOCAL_CATALOG_STORE_ID:
+            catalog_store.set_validators(catalog_store_validators)
+        # It is added to the sample catalog when it is setup initially
+        elif (
+            uri != SampleCatalog.SAMPLE_CATALOG_ID
+            and StoreURI.from_string(uri).type != StoreType.REMOTE_HTTP
```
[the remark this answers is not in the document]
sei-aderr: That works, pushed a fix.

test/store/cross_validators/test_cross_validators.py
```diff
@@ -136,5 +136,5 @@ def test_catalog_cross_validators(
         entry.quality_attribute = "not a quality attribute"
         with pytest.raises(RuntimeError):
             local_catalog_session.entry_mapper.create_with_header(
-                entry, INVALID_USER
+                entry, VALID_USER
```
sei-secheverria: This is still stomething strange here. create_with_header has the following signature:
```
    def create_with_header(
        self,
        entry: CatalogEntry,
        context: Any = None,
        user: typing.Optional[str] = None,
```

In all the calls made in this test function (not only this one), two unnamed params are passed. The first one is pretty clearly a CatalogEntry, since it is called entry. The second param passed is always a username (VALID or INVALID_USER). However, in the method signature, the second param is "context". So the user is being passed as context. Please fix by explicitly stating as the params ```(entry, user=INVALID_USER)```, or equivalent in the other calls in this function/file.

mlte/store/validators/cross_validator.py
```diff
@@ -0,0 +1,33 @@
...
+        user_store: Optional[UserStore] = None,
+        custom_list_store: Optional[CustomListStore] = None,
+    ):
+        """
+        Initialize a CrossValidator instance.
+        :param user_store: Catalog store to use for validation.
+        :param custom_list_store: Custom list store store to use for validation.
+        """
+        self.user_store = user_store
+        self.custom_list_store = custom_list_store
```
sei-secheverria: I still wonder if this has to be at this level. Why would the generic CrossValidator have to know about the specific stores? It is a way to already set up the stores  that are needed in the base constructor, but I am not fully convinced this is the best way to go.

I think it may be cleaner to have this at the specific constructor for the derived CrossValidator that needs it. For example, the ArtifactUserValidator only needs the UserStore, the ArtifactCustomListValidator only needs the CustomListStore, and so forth. We don't seem to be gaining much by having the common constructor, not even ensuring there will always be a store of the expected type as they can be None. The only benefit is kinda sorta defining the available stores. But even then, it is kinda clunky now: why are we adding only 2 of the 4 stores? Just because the specific CrossValidators we have need those two, but why should this base class/interface have to know that? Another benefit is defining the attribute for each store only once, and doing that set only once, but that is such a small benefit/reduction of code duplication that I don't think it's worth it, especially with the design implications it has as described above.

I'd recommend instead:

1. Remove the constructor from CrossValidator, making it a true interface.
2. Modify the specific cross validators to store the store they are receiving in an attribute, rather than calling super()
3.  Actually, the CrossValidator class can then be moved back to composite_validator.py, as there will be no circular imports.
4. As an aside, I'd rename the composite_validator module, assuming the CrossValidator class goes back in there. I think cross_validator.py is more appropriate, the Composite one is just an artifact to put several CrossValidators together.

mlte/store/validators/cross_validators.py
sei-secheverria: I am not sure we gain much by having all these cross_validators in one class. This file will continue growing more and more, in theory. Instead, I'd create a package/folder here called cross_validators/, and then separate each validator into a file/module with the name of each specific validator class. It is a bit more painful to handle more files, but easier to scale/maintain in the long run.

mlte/store/validators/cross_validator.py
sei-aderr: I wanted to have as much as possible defined in the interface, but agree that it ended up being quite clunky at the end. All of theses changes should be addressed

sei-secheverria (approved):

PR #821, opened 2025-12-03
Feature/gpu power 731

### LINKED ISSUE #731: Add GPU Power Utilization Measurement
Add a measurement that captures information about GPU power utilization for NVIDIA CUDA-enabled GPUs. Would probably be modeled very similarly to the GPU memory utilization one, sharing the underlying approach of using pynvml through _get_nvml_memory_usage_bytes. Probably a common class would need to be created, that presents methods to get either the memory utilization or power, and that consolidates common code that would be needed in both cases.

sei-amellinger:
I extracted the pynvml code to a new utils file.  I wasn't sure the naming convention for internal files/functions (such as prefix with _) so I just used normal naming conventions. I reworked the measurement code and the tests to use the new pynvml code base. Then added the new gpu power measurement code. I also updated the docs.

This code only supports one GPU per measurement, so the next step will be to do the multi-gpu min/max/average.

Resolves #731

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.3.0
+current_version = 2.4.0
 
 [bumpversion:file:./README.md]
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -63,6 +63,8 @@ jobs:
         python-version: ${{ fromJSON(needs.define-matrix.outputs.py_ver_matrix) }}
     runs-on: ${{ matrix.os }}
     steps:
+      - name: Clear space from tools folder, not needed because this runs in docker
+        run: rm -rf /opt/hostedtoolcache
       - uses: actions/checkout@v4
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v5
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 `MLTE` (pronounced "melt") is a framework and infrastructure for evaluating machine learning models and systems. To get started with the `MLTE` Python package, continuing reading below. The `MLTE` framework can be found in the <a href="https://mlte.readthedocs.io/en/latest/" target="_blank">documentation</a>, along with a more in-depth guide to <a href="https://mlte.readthedocs.io/en/latest/using_mlte/" target="_blank">using `MLTE`</a> that expands on the quick start guide below. For examples of use cases, see the <a href="https://github.com/mlte-team/mlte/tree/master/demo" target="_blank">demo folder</a>. 
 
-![Version Badge](https://img.shields.io/badge/release-v2.3.0-e19b38)
+![Version Badge](https://img.shields.io/badge/release-v2.4.0-e19b38)
 [![Python](https://img.shields.io/pypi/pyversions/mlte.svg)](https://badge.fury.io/py/mlte)
 [![Code Style: black](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/psf/black)
 [![Tests](https://github.com/mlte-team/mlte/actions/workflows/ci.yaml/badge.svg)](https://github.com/mlte-team/mlte/actions/workflows/ci.yaml)
--- a/demo/simple/0_negotiation.ipynb
+++ b/demo/simple/0_negotiation.ipynb
@@ -159,7 +159,7 @@
     "            measure=\"with an accuracy of 98%\",\n",
     "        ),\n",
     "        QASDescriptor(\n",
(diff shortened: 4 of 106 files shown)


docs/docs/mlte_measurements.md
```diff
@@ -13,7 +13,13 @@ The following are specific built-in measurement classes provided by MLTE:
...
     - Measures the amount memory being used per-GPU during an experiment.
     - Returns custom evidence in the form of an NvidiaGPUMemoryStatistics object.
-    - Requires a CUDA-capable NVIDIA GPU to be used, as well as the CUDA Toolkit (including the NVML library) to be installed. 
+    - Requires a CUDA-capable NVIDIA GPU to be used, as well as the CUDA Toolkit and the NVML library.
+    - Requires the pynvml library (`pip install nvidia-ml-py`) to be installed.
+- NvidiaGPUPowerUtilization
+    - Measures the power (in watts) used by the specific NVIDIA GPU.
+    - Returns custom evidence in the form of an NvidiaGPUPowerStatistics object.
+    - Requires a CUDA-capable NVIDIA GPU to be used, as well as the CUDA Toolkit and the NVML library.
+    - Requires the pynvml library (`pip install nvidia-ml-py`) to be installed.
```
sei-secheverria: Same here

mlte/measurement/pynvml_utils.py
```diff
@@ -0,0 +1,176 @@
...
+
+def aggregate_measurements_from_process(
+    pid: int,
+    poll_interval: float,
+    gpu_id: int,
+    fn: Callable[[types.ModuleType, int], float],
+    default: float = 0.0,
+) -> Tuple[float, float, float]:
+    """
+    :param pid: The process identifier
```
sei-secheverria: Maybe add a short line describing the purposes of the function

mlte/measurement/pynvml_utils.py
```diff
@@ -0,0 +1,176 @@
...
+
+    The callable is provided with a loaded pynvml as the first argument and the handle
+    to the device as a second argument, so that a call can be made directly to pynvml with
+    the handle. For example:
+
+    fn(pynvml, handle) -> float:
+        # NOTE: Pynvml exposes version 1 of the structure which has:
+        # total, used, free
+        memory_info = pynvml.nvmlDeviceGetMemoryInfo(handle)
+        return float(memory_info.used)
```
sei-secheverria: Is this still valid, for any fn that is received? Or was this only from the previous implementation, where only memory was obtained?

I think this part of the comment should be moved, if it is still needed, to the _get_nvml_memory_usage_bytes function in the corresponding module

mlte/measurement/power/nvidia_gpu_power_utilization.py
```diff
@@ -0,0 +1,149 @@
...
+        Monitor memory usage on a specific gpu.
+
+        :param pid: The process identifier
+        :param poll_interval: The poll interval, in seconds
+        :param unit: The unit to return the memory size in, defaults to statistics default unit.
+        :return: The captured statistics
+        """
+
+        # Keep collecting stats until the controlling process goes away.
+        # It might actually take the controlling process a while to start up the memory utilization
```
sei-secheverria: This seems to be outdated, is talking about memory utilization

test/measurement/power/test_nvidia_gpu_power_utilization.py
```diff
@@ -0,0 +1,354 @@
...
+@pytest.mark.skipif(
+    not has_torch_cuda(),
+    reason="NvidiaGPUPowerUtilization requires cuda to test.",
+)
+def test_power_validate_success() -> None:
+    m = NvidiaGPUPowerUtilization("identifier")
+
+    # Blocks until process exit
+    delay = 2
+    # TODO: Why specify units?
```
sei-secheverria: Because the NvidiaGPUPowerUtilization measurement expects a unit argument. But it also has a default value, so you shouldn't really need to specify units in this test case.

test/measurement/memory/test_nvidia_gpu_memory_utilization.py
```diff
@@ -55,35 +54,37 @@ def get_cuda_load_command(delay_sec: int = 2) -> list[str]:
...
-def has_pynvml():
-    """
-    Checks to see if pynvml is available without actually trying to import.
-    :return: True if available for import.
-    """
-    import importlib.util
-
-    spec = importlib.util.find_spec("pynvml")
-    return spec is not None
+# def has_torch_cuda() -> bool:
```
sei-secheverria: I think we can remove all this commented out text, as it has been moved to the test_pynvml_utils module

sei-secheverria (changes requested):
Looks very good, some minor things to adjust below.

PR #840, opened 2026-01-23
Separate CI commands into front/backend so they can be ran separately

### LINKED ISSUE #777: Set up QA make commands for frontend to run in a container
Currently, setting up the frontend Node environment is painful and usually not needed. To make it optional, change make commands that interact with the frontend QA actions to run inside an isolated container with the Node.js environment.

Some notes:
 - One of the targets inside ``docker/Dockerfile.frontend``, the one called ``frontend-env``, has the whole frontend's Node.js environment set up. This target could be used as the base for any QA commands.
 - Review how this would work with people that do want to set up a local env. Would there be an option to choose between them in some way?
 - Not directly related, but the ``docker/Dockerfile.python`` image could be used to do something similar for the Python side of the code, to run qa, though the dev dependencies would need to be installed.

sei-aderr:
Partially resolves #777

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 # Automation of various common tasks
 
 # -----------------------------------------------------------------------------
-# Python env setup.
+# Python env setup
 # -----------------------------------------------------------------------------
 
 .PHONY: venv-clean
@@ -30,7 +30,7 @@ check-schema:
 	poetry run python tools/schema.py vet mlte --verbose
 
 # -----------------------------------------------------------------------------
-# Doc building/checking.
+# Doc building/checking
 # -----------------------------------------------------------------------------
 
 # Doc generation.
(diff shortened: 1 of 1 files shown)


sei-aderr:
Will add on the other parts of the task to this branch

Makefile
```diff
@@ -94,15 +98,19 @@ demo-clean:
...
 qa-python: schema isort format lint typecheck demo-clean docs build-sample-catalog
 
-# Check all QA tasks for Python.
+# Check all QA tasks for Python
 .PHONY: check-qa-python
 check-qa-python: check-schema check-isort check-format lint typecheck docs check-sample-catalog
 
+# CI for Python bits
+.PHONY: ci-python
+ci-python: python-env-clean check-qa-python test demo-test
```
sei-secheverria, line 107: I now you closed this PR, I assume because you want to add more things first? But a small comment here: this ci-python command is missing a call to "venv" after python-env-clean, or the python env won't be recreated., similarly to what the ci-frontend does. You may also want to rename venv to python-venv or similar, for consistency.

PR #849, opened 2026-02-26
Container QA/CI

### LINKED ISSUE #777: Set up QA make commands for frontend to run in a container

sei-aderr:
Resolves #777 

Not perfect, but working. Scripts `run_frontend_qa.sh` and `run_python_qa.sh` are available to run either frontend or backend QA/CI. They will run the QA by default, but can pass the ci make commands (or technically any make command) a parameters to the script to do those instead.

Backend one does take forever to run, particularly the docs generation which I think is because it is volume mounted and it slows it down a lot. Not sure what a solution would be but could be looked into in the future.

Will add some docs for these scripts to the readme shortly


Makefile
```diff
@@ -1,21 +1,21 @@
...
-venv:
+.PHONY: python-venv
+python-venv:
 	python -m venv .venv && \
 	poetry lock && \
 	poetry install --with dev,demo --all-extras
 
 .PHONY: venv-redo
-venv-redo: venv-clean venv
+venv-redo: venv-clean python-venv
```
sei-secheverria: It is good that we added the "python-env-clean" below for consistency with the frontend one. However, now that we changed the venv command to python-venv, the venv-clean command is inconsistent. Maybe venv-clean should be renamed to venv-remove, since that is what it is doing? In theory, the current venv-clean may be more equivalent to the current frontend-clean in that the whole env is deleted, but when we are running the CI commands, we don't want to redo the python venv, but we do want to do it for the frontend, so it is not really equivalent as they are named right now.

Maybe what we should do for consistency is:
 - Rename venv-clean into python-venv-remove
 - Rename frontend-clean to frontend-env-remove

Makefile
```diff
@@ -85,24 +85,43 @@ typecheck:
...
 	cd demo && bash clean_all_nbs.sh simple GardenBuddy ReviewPro GradientClimber
 
-# QA for Python bits.
+# QA for Python bits
 .PHONY: qa-python
 qa-python: schema isort format lint typecheck demo-clean docs build-sample-catalog
 
-# Check all QA tasks for Python.
+# QA for Python bits, ran within a docker container
+.PHONY: qa-python-docker
```
sei-secheverria, line 103: Is there a reason there is this command for qa and ci for Python in Docker, but not for qa and ci of the frontend in Docker?

docs/docs/development.md
```diff
@@ -107,6 +107,15 @@ There are a couple of shorthand commands in the Makefile to run several of the b
 
 `make qa` and `make test` should be ran before every push to ensure that the changes made adhere to the QA standards and will pass the unit tests. These two encapsulate all of the commands below that are generally applicable.
 
+If you want to run either the frontend or backend QA/CI without setting up the environment, there are scripts available to run them within containers. These will mount your local version as a volume and run the QA/CI with the environment in the container while applying the results to your local copy.
+
+```bash
+$ cd docker && bash run_python_qa.sh # Python QA
+$ cd docker && bash run_python_qa.sh ci-python # Python CI
+$ cd docker && bash run_frontend_qa.sh # Frontend QA
+$ cd docker && bash run_frontend_qa.sh ci-frontend # Frontend CI
```
sei-secheverria: Do we need to expose this by calling the script directly? If we had docker commands in the Makefile, as we do for the python side (I think we are missing the frontend commands), we could just call here make *command*, which would be more consistent with what the rest of these docs are doing.

sei-secheverria (changes requested):
Looks great, but we should create a low priority ticket to look into the slowness for later. Added a few minor comments.

Makefile
sei-aderr: Works for me, adjusted this

Makefile
sei-aderr: Just forgot to add them

docs/docs/development.md
sei-aderr: Yeah, using the make commands with them all added is better

sei-aderr:
Added #850 for continued work on this

Makefile
```diff
@@ -151,15 +151,28 @@ typecheck-frontend:
 
 # QA for frontend (node.js) bits
 .PHONY: qa-frontend
-qa-frontend: lint-frontend typecheck-frontend
+qa-frontend: 
+	lint-frontend typecheck-frontend
```
sei-secheverria: There is a syntactical bug here. Dependencies (other targets) have to be in the same line after the colon, not in a new line. Things in a new line are executed as shell commands. This is currently failing due to this.

Makefile
```diff
@@ -151,15 +151,28 @@ typecheck-frontend:
...
+# QA for the frontend (node.js) bits, ran within a docker container
+.PHONY: qa-frontend-docker
+qa-frontend-docker:
+	cd docker && bash run_frontend_qa.sh qa-frontend
 
 # Check all QA tasks for frontend
 .PHONY: check-qa-frontend
-check-qa-frontend: check-lint-frontend typecheck-frontend
+check-qa-frontend:
+	check-lint-frontend typecheck-frontend
```
sei-secheverria: Same bug here

Makefile
```diff
@@ -151,15 +151,28 @@ typecheck-frontend:
...
 
 # CI for frontend (node.js) bits
 .PHONY: ci-frontend
-ci-frontend: frontend-env-clean frontend-env check-qa-frontend
+ci-frontend:
+	frontend-env-remove frontend-env check-qa-frontend
```
sei-secheverria: And here

sei-secheverria (changes requested):
Several commands are failing due to a newline issue, as described below.

Makefile
```diff
@@ -130,14 +149,28 @@ check-lint-frontend:
...
 
-# QA for frontend (node.js) bits.
+# QA for frontend (node.js) bits
 .PHONY: qa-frontend
 qa-frontend: lint-frontend typecheck-frontend
 
+# QA for the frontend (node.js) bits, ran within a docker container
+.PHONY: qa-frontend-docker
+qa-frontend-docker:
+	cd docker && bash run_frontend_qa.sh qa-frontend
```
sei-secheverria: This command is currently failing for me, for some reason. Not sure if it is an absolute error, or a problem with my local env and frontend files are being accidentally copied to the docker container, but the process stops when trying to run the frontend linter with this error:

```
cd mlte/frontend/nuxt-app && npx eslint --fix .

Oops! Something went wrong! :(

ESLint: 9.28.0

Error: Failed to load native binding
    at Object.<anonymous> (/mnt/app/mlte/frontend/nuxt-app/node_modules/unrs-resolver/index.js:372:11)
    at Module._compile (node:internal/modules/cjs/loader:1730:14)
    at Object..js (node:internal/modules/cjs/loader:1895:10)
    at Module.load (node:internal/modules/cjs/loader:1465:32)
    at Function._load (node:internal/modules/cjs/loader:1282:12)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:235:24)
    at cjsLoader (node:internal/modules/esm/translators:266:5)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/translators:200:7)
    at ModuleJob.run (node:internal/modules/esm/module_job:274:25)
```

sei-secheverria (changes requested):
Syntax errors were fixed. Python docker commands seem to be working, but qa frontend command is finding some errors when run inside the docker container.

Makefile
sei-aderr: So this does work on my mac. Could you just double check you have the updated `docker/run_frontend_qa.sh` file with the `--tmpfs /mnt/app/node_modules \` line and see if it errors again?

This was an error I was troubleshooting that led me to having to add it like that. I believe that yes, it is because the node modules from your local machine are being copied in that having them mounted a a tmpfs should have fixed. It was a circle of troubleshooting getting certain things to see the node modules, try and get other ones to ignore them, etc. I can try and find another solution if it still doesn't work for you

sei-secheverria: The sh file did have the --tmpfs command. But I am still seeing the same error, even though I've recreated by local venv a couple of times for unrelated reasons. 

I am not totally sure what the tmpfs arg is doing. But more in general, I am not sure I agree with the way the scripts are working now.  From what I see, they are mounting a lot of local folders. That is nice in terms that it is quick and it links to what you have locally, but on the other hand, half of the docker isolation is lost if we mount things and accidentally end up mounting things that interfere wit the test. I guess mounting instead of copying is slightly faster and does not require recompilation, but I think we will never get a good workaround with this. It may be the case that for the run_python_qa.sh script, mounting each subfolder is fine, since the python environment itself is in a separate, .venv folder that is not mounted. However, for the frontend one, since it has a mess of nested folders that are part of the environment, I think the best way is to simply copy the frontend code that actually has to be there, and not the code that does not have to be there, and then run the commands. Otherwise we'll always come back to this potential conflicts with local environments. 

Btw, I am not sure why we are even mounting things. mlte-frontend-env, which is the base container for the frontend qa one, is already copying exactly and only the files that are needed. Why is the script mounting again the code that we just copied over, and more folders that are not needed and are apparently the cause of the issues?

And discussing this, the same applies to the Python qa container. It already has the code from the mlte-python one. Why do we need to mount anything?

sei-aderr: I think that I agree it may be better to do sub pieces of the frontend to avoid the hack for the node modules.

The reason that I decided to mount everything is so that the qa fixes could be applied to the local environment. So if there are formatting errors, you can just run the QA container and it will make the changes in your local version. Did you have in mind that this would be mainly focused on qa-check/ci and just wanted it to be able to identify issues, or did you have another idea for how we could get those fixes out of the container?

sei-secheverria: I get that idea of applying the qa fixes to the local environment, but I think that, in the end, you can't really doing that properly when using Docker. By putting everything in docker you are changing the base environment, and your local environment may not work properly on that different base environment. I guess I was thinking more about qa-check and ci, because you are right, without mounting you won't get those benefits.

Note that there are two things: the local env, and the local files. In no case I think it is a good idea to mount the local env for the reasons above. For python, we are not mounting the local env (the .venv), and that is fine. But for the frontend we seem to be doing that, with node_modules and others, since that env is distributed. I agree that getting the qa fixes on your folder would be a benefit. To achieve that, you would have to mount parts of the frontend more piecemeal. You can give that a try to see if it works and we get the benefits of running the qa there.

sei-aderr: Alright take another look at it now and see. I made a couple changes + optimizations

- Renamed all files to `ops` instead of QA
- Removed default CMD from dockerfiles so it is clearer what the param is required and you have to supply the make command
- Made dev dep poetry install use `--no-root` so it can be saved as an unchanging layer, then install MLTE with poetry at the end of the python dockerfile
- Adjust frontend mount to mount all of the directories that would be looked at by QA/CI to avoid node modules issues and weird mount exceptions
- Unclear what exactly changed,  but mkdocs in the python QA now runs in ~30 seconds within the container when it used to take ~5min

docker/Dockerfile.frontend_ops
```diff
@@ -1,6 +1,10 @@
 # Image definition for MLTE frontend container.
 FROM mlte-frontend-env
 
+# Run nuxi prepare to generate .nuxt folder that eslint relies on
+WORKDIR /mnt/app/mlte/frontend/nuxt-app
+RUN npx nuxi prepare
```
sei-secheverria, line 6: This may be ok, but just wondering, why is this needed in this case, and why is this not needed in the regular Docker.frontend file, that prepares everything for the frontend?

docker/Dockerfile.python_ops
```diff
@@ -19,6 +19,9 @@ COPY .git .git
 COPY Makefile .
 COPY .flake8 .
```
sei-secheverria, line 21: Do note that in the .dockerfile in the root of the MLTE repo, Makefile and .flake8 are explicitly put as ignored. I guess it depends on how this is built from, that root dockerignore may be used or not, but for consistency, I'd remove these two files from the root dockerignore, to ensure they will be available when needed, which is the case here.

sei-secheverria (changes requested):
Looks good, and it is working now.  A very minor question and change below.

docker/Dockerfile.python_ops
sei-aderr: Each of these ops dockerfiles has its own dockerignore that they use with matching names. The python one needs to have its own dockerignore because it needs to have several things that are ignored in the base such as tests dir, tools dir, and these two files. 

I will remove these two from the main dockerignore, and then the frontend ops dockerfile would not need its own dockerignore, as it is the exact same besides Makefile not being ignored.

docker/Dockerfile.frontend_ops
sei-aderr: While `npm run build` is in the main dockerfile and partly generates the `.nuxt` dir, this doesn't generate `.nuxt/eslint.config.mjs` that eslint relies on to run which is part of our QA. It looks to only run when `npx nuxi prepare` is ran, which is a simplified dev setup that doesn't create the prod resources like `npm run build` does. `.nuxt/eslint.config.mjs` is should also created when you do `npm run dev` but our `frontend-build` doesn't do that, it would happen within the second stage, but that is separate and only contains `.output` to be slim.

docker/Dockerfile.python_ops
sei-secheverria: That makes sense, and simplifying things this way makes it a bit more tractable, thanks!

sei-secheverria (approved):

Issue #843, opened 2026-02-16
Exporting Artifact Store Dependencies Automatically


sei-secheverria:
Part of #680. Whenever something is exported, all dependencies should be exported as well, even if not explicitly requested. So all CustomLists used in a NegotiationCard should be exported alongside it, for example; same with users. Probably a good idea to somehow hang on to the current validators to take advantage of the dependency details that are implied there.

PR #851, opened 2026-03-03
Feature/store export

### LINKED ISSUE #844: Basic Export Functionality
Export functionality should allow to select things in a tree in this way and at this granularity (everything could be selected completely, or only some sub items).

- Artifact Store
  - Specific Model(s)
    - Specific Version(s)
- CustomList Store
  - Specific List(s)?
- User Store
  - Specific User(s)?
- Default Test Catalog Store

Export function should receive a structure that mirrors the one above indicating which parts to export. It can then create a temporary folder, copy the corresponding resources there, and create a zip/tar file with the result, ready to be used. Library functionality should just copy this zip/tar to a provided folder.

### LINKED ISSUE #874: Change Data Schema Tag
In the Negotiation Card, under Data there is Data Schema, and then when you want to add a field it says Data Scema , Data Schema 2, .... This seemed to cause confusion at our last demo given that it seems to imply that there is more than one data schema. I agree. Would it be possible to change it so that it says Field 1, Field 2, ...?

sei-aderr:
Resolves #844, resolves #874

Has an initial implementation of the library export. Let me know if we want anything added in for this first version

Questions
1. Custom list names are based on the `CustomListName` enum, so right now they have to be specified when calling export as `CustomListName.CLASSIFICATION`. Do you think that is too not user friendly?
2. Did you have thoughts on how to initially handle permissions & groups for users? Easiest would be keep all, or delete all but each of those paths could have their own issues.


mlte/store/export/export.py
```diff
@@ -0,0 +1,105 @@
...
+
+from mlte.custom_list.custom_list_names import CustomListName
+from mlte.session.session import session
+from mlte.store.artifact.store_session import ManagedArtifactSession
+from mlte.store.catalog.store_session import ManagedCatalogSession
+from mlte.store.custom_list.store_session import ManagedCustomListSession
+from mlte.store.user.store_session import ManagedUserSession
+
+
+class ExportSpec(TypedDict):
```
sei-secheverria: A comment explaining this class would be nice.

mlte/store/export/export.py
```diff
@@ -0,0 +1,105 @@
...
+    models: dict[str, list[str]]
+    custom_lists: list[CustomListName]
+    users: list[str]
```
sei-secheverria: I'd also add comments to each of this attributes for consistency of how we are documenting classes such as this one

sei-secheverria: We should also be able to say if we want to export the default catalog or not, at the very least.

mlte/store/export/export.py
```diff
@@ -0,0 +1,105 @@
...
+        "custom_lists": {},
+        "users": [],
+        "default_test_catalog": [],
+    }
+
+    # Read items to be exported
+    with ManagedArtifactSession(
+        session().stores.artifact_store.session()
+    ) as artifact_store:
+        if "models" in export_spec:
```
sei-secheverria: Too many floating strings... better to define constants for these keys, and reuse them where needed.

mlte/store/export/export.py
```diff
@@ -0,0 +1,105 @@
...
+            for model_id in export_spec["models"]:
+                export_dict["models"][model_id] = {}
+                for version_id in export_spec["models"][model_id]:
```
sei-secheverria: What happens if I just specify a model, and I want all versions of the model? Nothing would be exported, I think. I believe this code currently only supports specifying specific versions to export. We should also support exporting all versions of a model by just stating the model name and no version names, as well as all models by just stating no model name, but you have to review exactly what that means, and how to differentiate between all models and no models.

mlte/store/export/export.py
```diff
@@ -0,0 +1,105 @@
...
+                        (model_id, version_id)
+                    ):
+                        export_dict["models"][model_id][version_id][
+                            artifact_id
+                        ] = artifact_store.artifact_mapper.read(
+                            artifact_id, (model_id, version_id)
+                        ).to_json()
+
+    with ManagedCustomListSession(
+        session().stores.custom_list_store.session()
```
sei-secheverria: I have the impression that this function will keep on growing substantially, if not now, when we add dependency handling and such later. I'd recommend splitting it into subfunctions for exporting each store, and calling each function by the main export function. Thinking that later we will also want to import, we may want to group import/export of each store into separate classes as well.

mlte/store/export/export.py
```diff
@@ -0,0 +1,105 @@
...
+    ) as custom_list_store:
+        # TODO: This requires the input to be things like `CustomListName.CLASSIFICATION` in the spec, consider if we want this
```
sei-secheverria: I think it is fine

mlte/store/export/export.py
```diff
@@ -0,0 +1,105 @@
...
+        if "custom_lists" in export_spec:
+            for custom_list_id in export_spec["custom_lists"]:
```
sei-secheverria: Well, same comment for all stores, how to say whether you want all lists or none, not only none or a specific list of lists.

sei-secheverria:
> Resolves #844
> 
> Has an initial implementation of the library export. Let me know if we want anything added in for this first version
> 
> Questions
> 
> 1. Custom list names are based on the `CustomListName` enum, so right now they have to be specified when calling export as `CustomListName.CLASSIFICATION`. Do you think that is too not user friendly?
> 2. Did you have thoughts on how to initially handle permissions & groups for users? Easiest would be keep all, or delete all but each of those paths could have their own issues.

1. I think that is fine.
2. Comments in the review.

mlte/store/export/export.py
```diff
@@ -0,0 +1,105 @@
...
+                for (
+                    custom_list_entry
+                ) in custom_list_store.custom_list_entry_mapper.list_details(
+                    custom_list_id
+                ):
+                    export_dict["custom_lists"][custom_list_id].append(
+                        custom_list_entry.to_json()
+                    )
+
+    # TODO: Handle permissions & groups
```
sei-secheverria, line 162: Even though this is internal to the user store, this basically feeds into #843. The same logic should be applied in this case. But we can ignore that for this first ticket, and then tackle it over there. This case will be a bit simpler since the dependencies are in the same store, so finding where the dependency is is trivial, but the same logic about flags to import overwritting, not, etc, should be used.

mlte/store/export/export.py
```diff
@@ -0,0 +1,105 @@
...
+        session().stores.user_store.session()
+    ) as user_store:
+        if "users" in export_spec:
+            for user in export_spec["users"]:
+                export_dict["users"].append(
+                    user_store.user_mapper.read(user).to_json()
+                )
+
+    with ManagedCatalogSession(
+        session().stores.catalog_stores.catalogs["local"].session()
```
sei-secheverria: "local" is a constant defined somewhere, use that constant instead of manually copying the string here

mlte/store/export/export.py
```diff
@@ -0,0 +1,105 @@
...
+
+class ExportSpec(TypedDict):
+    models: dict[str, list[str]]
+    custom_lists: list[CustomListName]
+    users: list[str]
+
+
+def library_export(export_spec: ExportSpec, output_path: Path) -> None:
+    """
+    Export when using the library, writes the exported JSON to output_path as zip file.
```
sei-secheverria: Looks good, but I would not call this anything about using a library. I'd call it export_to_file. We may end up using this from the frontend as well. In any case, there is no need to put in the name/comments who will use this.

mlte/store/export/export.py
sei-secheverria, line 1: Note that we are also missing unit tests.

sei-secheverria (changes requested):
Looks good overall, several things to review below.

mlte/store/export/export.py
sei-aderr: Yes, the plan is to handle it in a future PR. Just wanted to have a note

sei-aderr:
Ok, all comments besides unit tests are addressed. Working on those now.

However, a "bug" that I noticed is that because of the way we set up cards to be stored at the model level, not the version level but are also part of each version, the current export implementation will add the card into every version of the model when exported. While not technically incorrect, would either have to be handled in the export or import in some way.

mlte/store/export/export.py
```diff
@@ -0,0 +1,185 @@
...
+    """Dict of models to be exported. Key is model ID, value is list of versions."""
+
+    custom_lists: list[CustomListName] | str
+    """List of custom lists to be exported."""
+
+    users: list[str] | str
+    """List of user IDs for users to be exported."""
+
+
+def session_export(export_spec: ExportSpec, output_path: Path) -> None:
```
sei-aderr, line 40: Working through a couple ideas of what would be the best way to handle this now. This works, but is not my favorite because if we want to have non-file exports, we would need a wrapper method for each of those too.

I am thinking it may make sense to have a function like this that can be called by `export_to_file` to get whichever stores are active. Then any similar export that isn't to file, could call the same thing and then pass those stores down to each object type of export.
```
    if session().stores.artifact_store:
        return session().stores

    # Note, this has to look at one particular kind of store, because by importing state to look at it, the global state is setup. 
    if state.stores._artifact_store:
        return state.stores
```

I initially thought this could make testing it a pain, but as I think about it more I do not think so. We could have one test to call it with session_stores setup, and one that calls it after setting up state_stores. Just want to confirm that approach seems viable with the changes you are in the process of making to the test setup.

sei-secheverria:
> Ok, all comments besides unit tests are addressed. Working on those now.
> 
> However, a "bug" that I noticed is that because of the way we set up cards to be stored at the model level, not the version level but are also part of each version, the current export implementation will add the card into every version of the model when exported. While not technically incorrect, would either have to be handled in the export or import in some way.

Yes, we should discuss it. We could punt it to import if you want for now, but the solution may be to export it as they are stored in the FS store, which is at the model level, with some additional code/extra handling, though not sure about the best way to do this. Do note that the artifacts themselves do have an attribute, the "level" field in the header, that indicates whether they should live at the model or version level.

mlte/store/export/export.py
sei-secheverria: What do you mean by "non-file exports"? Exporting to some other format that is not a zip file? What would that be? Not sure I am following. I'd rather clarify that before discussing forward the ideas above.

sei-aderr: Only other one would just be the straight JSON that I was thinking of. But I suppose just returning the zip file from the endpoint to be downloaded is probably sufficient.

sei-secheverria: To simplify things, let's assume just the zipfile for now. It may be the case that later we'll end up merging the session and state things into roughly one (I think state should have a session object, which would at least simplify things a bit), so I wouldn't spend too much time on that for this first pass.

sei-secheverria: After delving deeper into this and reviewing the current state/session structure, I do not think that having a session_export() and a state_export() functions, and/or a way to detect which is being used, is the proper way to go. I believe that, even though this module should stay here as it is, the SessionStores (to be soon renamed into UnifiedStore) should have a method to export, that would call the methods here (basically, would call export_to_file), passing its stores as params.

I'd restructure this a bit to work this way:

 - SessionStores, add a method called export(export_spec, output_path, artifact_store, user_store, customlist_store, catalog_stores) that simply calls the export_to_file here. (It is a bit more painful to pass each store separately, but passing SessionStores directly would create a circular import. In non-python, this would be solved by an interface, but it is weird using interfaces in Python. If you can find a simple way to do that here, that is fine too, as long as we don't end up with a circular import, which won't run).
 - Then all knowledge about where the stores are is out of this module. When using this through the library, the user would have to get the current SessionStores and call it to export it, something like get_session().stores.export(....). The backend would get things from the state, like so: state.stores.export(...). In this way we are consistent with the rest of the system, where the entry point decides where the stores are, instead of us trying to guess it. And for the lib class, calling get_session().stores.export(...) instead of a simple export.export(...) still makes sense, since it is the current stores in the current sessions what we want to export.

sei-aderr: That works for me and sounds like a better solution than trying to figure out what kind of store is available for the export. 

When we talked before I didn't like having to pass all of the stores to the export, and I still don't that much but having to try and determine what kind of store are available is worse and not consistent with how anything else in the code works. 

Once you have all the changes ready in your PR, I can review that and then get this working on top of those changes

sei-aderr: Although this currently does cause a circular import. `session_stores` will now import `export` to have access to the export functionality, but `export` relies on `SessionStores.LOCAL_CATALOG_STORE_ID` to grab the local catalog in `_export_catalogs`.

So will have to re-arrange in some way to accomadate  for that

sei-secheverria: Ok, PR #856 updated with some latest refactorings, so you can review and use them.

Yes, I dislike passing all stores separately too. One solution would be to have a two-tiered thing, with one containing only the stores, and another the extra logic as well as the first one, and then passing only the first one to the export function. Another alternative would be to define an interface (like, store list), and have the current SessionStores implement that interface, and pass it as that interface to the export function. Neither of those two options work very well in Python; they could be made to work, but unless there are several cases where this would be reused, I don't think it is worth the effort.

sei-secheverria: > Although this currently does cause a circular import. `session_stores` will now import `export` to have access to the export functionality, but `export` relies on `SessionStores.LOCAL_CATALOG_STORE_ID` to grab the local catalog in `_export_catalogs`.
> 
> So will have to re-arrange in some way to accomadate for that

Sure, you'll need to find some way around that. Maybe moving those constants to a separate module.

sei-aderr:
Got everything updated ready for another review @sebastian-echeverria

mlte/store/export/export.py
```diff
@@ -0,0 +1,205 @@
...
+    # TODO: When getting multiple versions, this exports the card under each version. Should not do that
+    with ManagedArtifactSession(
+        artifact_store.session()
+    ) as artifact_store_session:
+        if export_spec.models is None:
+            return output_dict
+
+        if export_spec.models == {}:
+            for model_id in artifact_store_session.model_mapper.list():
+                export_spec.models[model_id] = []
```
sei-secheverria: I understand this is handling the issue of exporting all models by explicitly adding all models to the spec if no models are indicated in the spec. There are a couple of improvements we could do here:
 - Right now this is very mysterious. At least comments would be needed here to avoid having to figure out what is going on.
 - It is a bit inconsistent in terms of responsibilities for the export function to modify the spec. I would, at the very least, move this to be a function in the spec itself. The export function should not need to know this much about internal logic of what empty should be translated into. Not sure the best solution, but the quickest one would be to make this a method of spec that is simply called here
 -  A functionality issue I have here is: how do I differentiate "I don't want to import any models" vs "import all models"? This logic seems to imply now the opposite of what we had before, that I can't say I don't want any models at all. We should be able to specify both extremes, in some way.

mlte/store/export/export.py
```diff
@@ -0,0 +1,205 @@
...
+
+        for model_id in export_spec.models:
+            if export_spec.models[model_id] == []:
+                export_spec.models[model_id] = (
+                    artifact_store_session.version_mapper.list(model_id)
+                )
```
sei-secheverria: I assume this is the same but for the versions. The same concerns described above for the model apply.

mlte/store/export/export.py
```diff
@@ -0,0 +1,205 @@
...
+
+    with ManagedCustomListSession(
+        custom_list_store.session()
+    ) as custom_list_store_session:
+        if export_spec.custom_lists is None:
+            return output_dict
+
+        if export_spec.custom_lists == []:
+            for custom_list_id in CustomListName:
+                export_spec.custom_lists.append(custom_list_id)
```
sei-secheverria: Same comments as for model/version: comments, funciton, differentiate all from none.

mlte/store/export/export.py
```diff
@@ -0,0 +1,205 @@
...
+    """Return list of users specified in export_spec."""
+    output_dict: dict[str, Any] = {}
+
+    # TODO: Handle permissions & groups
+    with ManagedUserSession(user_store.session()) as user_store_session:
+        if export_spec.users is None:
+            return output_dict
+
+        if export_spec.users == []:
+            export_spec.users = user_store_session.user_mapper.list()
```
sei-secheverria: Same comments as for the two previous stores.

mlte/store/export/export.py
```diff
@@ -0,0 +1,205 @@
...
+    """Specification of MLTE store objects to be exported."""
+
+    models: dict[str, list[str]] | None = None
+    """Dict of models to be exported. Key is model ID, value is list of versions."""
+
+    custom_lists: list[CustomListName] | None = None
+    """List of custom lists to be exported."""
+
+    users: list[str] | None = None
+    """List of user IDs for users to be exported."""
```
sei-secheverria: One of the comments was ignored in the last review: we should also be able to say what catalogs we want to export. Right now I think we are only exporting the local catalog, but we should at least be able to say if we want or do not want to export it. Currently we are always exporting it.

mlte/store/export/export.py
```diff
@@ -0,0 +1,205 @@
...
+
+def _export_catalogs(catalog_stores: CatalogStoreGroup) -> dict[str, Any]:
+    """Return the local test catalog."""
+    output_dict: dict[str, Any] = {LOCAL_CATALOG_STORE_ID: []}
+
+    with ManagedCatalogSession(
+        catalog_stores.catalogs[LOCAL_CATALOG_STORE_ID].session()
+    ) as catalog_store_session:
+        for catalog_entry in catalog_store_session.entry_mapper.list_details():
+            output_dict[LOCAL_CATALOG_STORE_ID].append(catalog_entry.to_json())
```
sei-secheverria: I don't remember what we had discussed, but why do we only support exporting the local catalog? I get it that we don't want to export the sample nor remote ones, but in theory, there could be more than one local catalog, correct?

mlte/store/export/export.py
```diff
@@ -0,0 +1,205 @@
...
+
+    with zipfile.ZipFile(
+        os.path.join(output_path, EXPORT_ZIP_FILE), "w", zipfile.ZIP_DEFLATED
+    ) as zip_export_file:
+        zip_export_file.writestr(
+            EXPORT_JSON_FILE, json.dumps(export_json, indent=4)
+        )
+
+
+def _export(
```
sei-secheverria, line 162: I wonder if we should somehow manage the remote stores differently. We are kinda implicitly doing this for catalogs, but currently, the artifact store, the user store, and the custom lists could be remote stores. Do we want to allow exporting them if that is the case? Maybe the answer is yes, but we should discuss to ensure this is explcitly what we want.

test/store/export/conftest.py
```diff
@@ -0,0 +1,60 @@
...
+ALL_EXPORT_SPEC = ExportSpec(
+    models={},
+    custom_lists=[],
+    users=[],
+)
+
+ARTIFACT_EXPORT_DATA = "artifact"
+CUSTOM_LIST_EXPORT_DATA = "custom_list"
+USER_EXPORT_DATA = "user"
+CATALOG_EXPORT_DATA = "catalog"
```
sei-secheverria: Maybe these could be a slightly more complex JSON string, rather than just a string? It is confusing if it is so different from what you would actually would get. It does not have to be a full store string, but some short string that indicates that this is a JSON structure.

sei-secheverria (changes requested):
Looks good, but there are a couple of functionality issues, and some suggested design improvements, below.

mlte/store/export/export.py
sei-aderr: I agree that this wasn't my favorite implementation, but wasn't sure of a better way to do it. I got rid of the option of a string all character because it provided the option for the user to enter any string and run into issues. I tried it with a literal, but ran into lots of typecheck issues with the 3 separate type options, especially in the artifact with it nested. 

I could add it in the spec, but that does make the spec and the export both rely on receiving the stores. Might be able to add some extra handling there to make it clearer whats happening that can be isolated though so I'll give that a shot. Do you have any good ideas for a better way to specify some/none/all?

There is still the option to do no models, you would just pass the None value to the spec. So you either specify which ones explicitly, give an empty list/dict for all, or None for nothing. Which I should add tests for the None case to ensure it works speaking of that case.

test/store/export/conftest.py
sei-aderr: I had just done this because they are only used to place in the dict with `_export` and make sure they got placed in the right place in the dict. Plan was to just do this and in the future could adjust to fit what was needed at that time, I can add some basic structure though.

mlte/store/export/export.py
sei-aderr: I think that we would want to allow for remote stores of any type to be exported. Having them not be able to be exported would in some way break the idea of having a centralized MLTE instance. If you worked on a central instance for collaboration in some way, I think it is reasonable that you would be to be able to export anything there, even just to save a local copy or something similar. You would already have access to them (assuming we use permissions properly for this export, which is a separate thing I hadn't thought through), so I don't see a reason why you shouldn't be able to export.

Did you have a case that you would think you would't want it to be possible to export a remote store?

mlte/store/export/export.py
sei-aderr: You had it listed in the original ticket as just the default catalog store and I know we talked about it but don't remember why we landed on that. Should be simple enough to add the option for any store, so I will add that

mlte/store/export/export.py
sei-secheverria: Ok, I had not realized None could be used. So:
 - I think it is fine to have None/empty list/filled in list. However, this should be documented. Maybe in the Spec class, in the comment for each attribute? Or the main class comment?
 - I still think we should isolate that functionality and comment it as described above, but there is no need to change the functionality itself.

mlte/store/export/export.py
sei-secheverria: No, just wondering if it made sense. I agree that the simplest thing for now would be to just allow it. No changes needed for that.

test/store/export/conftest.py
sei-secheverria: I agree, I think the main issue is that it can be confusing to grasp at a quick pass. It could be enough to do something like ARTIFACT_EXPORT_DATA = "{model_1: "model_data"}" or whatever, to slighly simulate a store and make it clearer what is going on here. It doesn't have to be more exhaustive than that.

mlte/store/export/export.py
sei-aderr: Do you have any thoughts on whether an export of specific entries in a catalog would want to be exported, or just the entire catalog each time?

sei-secheverria: I'd assume entire catalog

sei-secheverria (approved):

PR #856, opened 2026-03-16
Feature/env credentials

### LINKED ISSUE #818: Allow library to set current user
Currently, when using a library, the user that is set when changing resources (such as writing artifacts) to a store is by default None, except when using an HTTP URI, in which case the user/password to be used in configured as part of the URI. While this works, this can be inconsistent (while not critical, one may want to have a specific user as the author of local artifacts) and insecure (in terms of the credentials for an HTTP store).

Change this so there is a way to set the current user, similar to how it can be done for the context and store URI (through a method and through env variables), in a way that is secure and allows a person using the library to, transparently, configure a user to be used for either local or remote HTTP store access.

### LINKED ISSUE #852: Add HTTP User Store
Initially, the HTTPUserStore was not implemented as it was not a common use case. However, with the integration of the stores, this is now needed to support any HTTP store, and to support cross validation in that case.

### LINKED ISSUE #853: Add unit tests to SessionStores
SessionStores does not have unit tests, which means some critical things are not properly tested. Add them, and while doing it, refactor the methods to create test stores so that they are more consistent, less duplicative, and can be reused more easily.

### LINKED ISSUE #855: Bug in TestCatalog Remote HTTP Store
The remote HTTP store has a bug that assumes that remote catalogs will have an intermediate id combination that should only be used for passing a request from a backend to another. Fix it so that this intermediate, combined format is only used internally, and in the initial request, when trying to get an entry from a remote store that is not the one being asked directly.

sei-secheverria:
- Session: Adds new feature to set current credentials when using library. For now, this is only used for giving user attribution to artifacts created/updated by the current user. Fixes #818
 - Session - Tests: Added unit tests for SessionStores and some other missing ones. Fixes #853
 - User Store: Implemented an underlying HTTP store for the UserStore. Fixes #852
 - Policies: In the case of the user policies, moved them into each of the underlying CRUD methods in the store's mappers, to ensure policies are applied regardless of whether the backend or library was creating a user.
 - Policies - Refactoring: Abstracts out setting of common policies for new users and models into separate modules, instead of embedded into backend endpoints, so they can also be executed by users and models created without the backend more easily.
 - Policies - Refactoring: Separated Policy object from PolicyStore abstraction to avoid circular dependencies
 - TestCatalog: Simplified HTTP catalog store to avoid having extra complicated steps for bundling the remote catalog id. Fixes bug due to this bundling complexity. Now the only strange thing is that, for accessing a remote catalog, besides indicating the remote catalog in the query param or catalog_id in the header, the local catalog has to be bundled as part of the entry id, as a prefix. Fixes #855
 - Test Catalog - Refactoring: Extracted our weird handling of remote catalog ids to separate modules for better isolation.
 - Users: Added better method for comparing users that takes into account groups may be in different orders
 - Testing Stores: Unified as much as possible test store handling methods. Reduced duplication by having common ones, ensured all test stores have same methods, and added centralized patches for store_setup that allows to set all stores to their test versions when needed.
 - Expanded use of conftest.py to avoid having weird import checks when importing fixtures
 - Fixed bugs in GPU measurement tests
 - Some minor bug fixes and deduplications


mlte/store/base.py
```diff
@@ -93,6 +93,11 @@ def from_string(uri: str) -> StoreURI:
             # If we got here, the stucture of the URI is fine, but the prefix is unknown.
             raise RuntimeError(f"Unsupported store URI: {uri}") from e
 
+    @staticmethod
+    def from_type(type: StoreType, path: str = "") -> StoreURI:
```
sei-aderr, line 97: It looks like this is only used in tests, does it make sense for it to be a part of the `StoreURI` class and not in a test fixture?

test/backend/api/endpoints/test_catalog_entry.py
```diff
@@ -337,16 +340,21 @@ def test_search(
...
+    test_entry = get_test_entry()
+    create_entry_using_admin(test_entry, test_api)
 
     url = get_entry_uri()
     res = test_client.post(f"{url}/search", json=Query().to_json())
     assert res.status_code == codes.OK
 
     collection = res.json()
-    assert len(collection) == 1
+    assert len(collection) >= 1
```
sei-aderr: Is `test_api_fixture` shared in a way that would result in there being more than one entry returned from the search? Looks like the next check just makes sure that the test entry is in the returned collection anywhere.

Do we want to make this test more in depth? It almost seems like a test that an item can be created, and we just use search to test it. It feels like we should specify what we are searching for to make the test of the search more direct.

test/store/utils.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+    Yield catalog store types.
+    :return: Store type.
+    """
+    for store_type in StoreType:
+        yield store_type
+
+
+def create_api_and_http_uri(
+    user: Optional[UserWithPassword] = None,
```
sei-aderr, line 24: Why does this method have this user param if it is just overwritten within the function no matter its value? It looks like all the `conftest.py` except catalog call it with a user as parameter.

mlte/artifact/artifact.py
```diff
@@ -93,20 +93,23 @@ def save(
         Save an artifact with parameters from the configured global session.
 
         This is equivalent to calling:
-            artifact.save_with(session().context, session().store)
+            artifact.save_with(get_session.context, get_session.store)
```
sei-aderr: Can you change this to `get_session()`

mlte/artifact/artifact.py
sei-aderr, line 90: Now that we are able to set a user using the library, do we want to continue allow objects to be written without a user? I think there are still cases where we do want to like creating the default custom list and catalogs, but those can be written by admin. Looks like catalogs already are, and custom lists don't save user info.

I don't have a super strong opinion either way at the moment, but we could discuss pros/cons once we both think through implications.

test/store/artifact/conftest.py
```diff
@@ -0,0 +1,121 @@
...
+
+# The mode identifier for default context
+FX_MODEL_ID = "model0"
+
+# The version identifier for default context
+FX_VERSION_ID = "v0"
+
+
+@pytest.fixture(scope="function")
+def store_with_context() -> Tuple[ArtifactStore, Context]:
```
sei-aderr: I think this should specify `artifact_store_with_context`, or something similar. Especially because it is being auto imported so many places by being in this `conftest` file it would be nice to be more clear exactly what you are getting when you call it.

test/store/artifact/conftest.py
```diff
@@ -0,0 +1,121 @@
...
+from mlte.store.artifact.underlying.http import HttpArtifactStore
+from mlte.store.artifact.underlying.memory import InMemoryStore
+from mlte.store.artifact.underlying.rdbs.store import RelationalDBArtifactStore
+from mlte.store.base import StoreType, StoreURI
+from mlte.user.model import UserWithPassword
+from test.store.defaults import IN_MEMORY_SQLITE_DB
+from test.store.utils import create_api_and_http_uri
+
+
+def _create_memory_store() -> InMemoryStore:
```
sei-aderr, line 28: It looks like this memory store and custom list do not use a cached memory version,  but catalog and user do. Was that an intentional choice?

test/store/user/conftest.py
```diff
@@ -0,0 +1,94 @@
...
+        create_user_store(
+            StoreURI.from_type(StoreType.LOCAL_FILESYSTEM, str(path))
+        ),
+    )
+
+
+def _create_rdbs_store() -> RelationalDBUserStore:
+    return RelationalDBUserStore(
+        uri=StoreURI.from_string(IN_MEMORY_SQLITE_DB),
+        connect_args={"check_same_thread": False},
```
sei-aderr: Why is this required for user, but not for any of the other store types?

mlte/backend/api/endpoints/user.py
```diff
@@ -86,31 +87,7 @@ def create_user(
...
-            # Give user permissions to modify its data.
-            own_user_policy = Policy(
-                ResourceType.USER, resource_id=user.username
-            )
-            own_user_policy.save_to_store(user_store)
-            own_user_policy.assign_to_user(user)
-
-            # Store the user.
-            new_user: BasicUser = user_store.user_mapper.create(user)
+            new_user = user_store.user_mapper.create(user)
```
sei-aderr, line 90: Should this also be type hinted? Looks like its a normal `User` now

mlte/backend/api/endpoints/user.py
```diff
@@ -137,15 +114,11 @@ def edit_user(
...
         try:
+            # TODO: this is a weird permission check that is outside the common permission checks.
             # We only want to allow admins to edit a user's groups.
             if current_user.role != RoleType.ADMIN:
-                # If not admin, keep current groups and ignore the new ones, if any.
-                current_groups = user_store.user_mapper.read(
-                    user.username
-                ).groups
-                user.groups = current_groups
+                user = user_policy.ignore_new_groups(user, user_store)
```
sei-aderr: Why do we have to do this instead of just skipping the edit?

mlte/store/catalog/underlying/http.py
```diff
@@ -1,20 +1,32 @@
 """
 Implementation of HTTP catalog store group.
+
+NOTE: indicating the remote catalog that an entry should be sent to/read from is not properly supported through
```
sei-aderr, line 4: 🙂

mlte/store/user/policy/policy_store.py
```diff
@@ -0,0 +1,77 @@
...
+
+from __future__ import annotations
+
+import mlte.store.error as errors
+from mlte.store.user.mappers import GroupMapper, PermissionMapper
+from mlte.store.user.policy import Policy
+
+
+class PolicyStore:
+    """Handles access to stored policies. It is not a store in the same sense as the rest, does not derive from Store."""
```
sei-aderr, line 11: Would be nice to differentiate this naming in some way if it is not an actual store, even something like `PolicySubStore` if it is just a subpiece of the user store

mlte/store/user/policy/policy_store.py
```diff
@@ -0,0 +1,77 @@
...
+        permissions: dict[str, bool] = {}
+        for group in policy.groups:
+            for permission in group.permissions:
+                # Store permissions in dict for later removal, to avoid trying to re-remove already deleted ones.
+                permissions[permission.to_str()] = True
+
+            self.group_mapper.delete(group.name)
+
+        # Now remove all permissions.
+        # TODO: note that this main leave other groups using these permissions dangling. Not trivial to check if
```
sei-aderr: I think this is a typo, `this (may) leave other groups dangling`?

mlte/store/user/underlying/rdbs/store.py
sei-aderr: Nitpick for consistency, could we change this be call `super()`? The other implementations do it that way.

sei-aderr: And while we are adjusting these store implementations, why does http not allow for the adjustment of the `add_default_data` param, and why does fs not call super?

I see that http is handled within the super, but I feel like it may make more sense for it to be more explicit within the http store implementation itself that it doesn't get the defaults

test/artifact/test_artifact.py
```diff
@@ -135,3 +130,89 @@ def test_artifact_parents(
...
+
+
+@pytest.mark.parametrize("artifact_type", ArtifactType)
+def test_save_explicit_user(
+    artifact_type: ArtifactType,
+    store_with_context: tuple[ArtifactStore, Context],
+):
+    """Test explicit user is properly set as creator."""
+    store, ctx = store_with_context
+    user = "explicit-user"
```
sei-aderr, line 161: Is the user name not checked to be a valid user in the the user store when the artifact is saved with `save_with`?

test/artifact/test_artifact.py
```diff
@@ -135,3 +130,89 @@ def test_artifact_parents(
...
+):
+    """Test that user from session is used when saving artifact."""
+    store, ctx = store_with_context
+    user = "session-user"
+    artifact_model = ArtifactModelFactory.make(artifact_type)
+    artifact = ArtifactFactory.from_model(artifact_model)
+
+    set_credentials(user, "password")
+    set_context(ctx.model, ctx.version)
+    get_session()._stores._artifact_store = store  # type: ignore
```
sei-aderr, line 187: What is this line doing here and the same line in the next test?

mlte/store/unified_store.py
```diff
@@ -126,12 +127,12 @@ def setup_stores(
...
-def _setup_arifact_store(stores_uri: str, stores: SessionStores) -> None:
+def _setup_arifact_store(stores_uri: StoreURI, stores: UnifiedStore) -> None:
     """
-    Creates an artifact store, validators for it, and adds it to stores
+    Creates an artifact store, validators for it, and adds it to stores.
 
-    :param stores_uri: The store URI string
-    :param stores: The SessionStores object that the new store will be added to
+    :param stores_uri: The store URI.
+    :param stores: The SessionStores object that the new store will be added to.
```
sei-aderr: `UnifiedStore` type

mlte/store/unified_store.py
```diff
@@ -152,13 +153,15 @@ def _setup_arifact_store(stores_uri: str, stores: SessionStores) -> None:
...
+    catalog_uris: dict[str, StoreURI],
 ) -> None:
     """
-    Creates catalog stores, validators for them, and adds it to stores
+    Creates catalog stores, validators for them, and adds it to stores.
 
-    :param stores_uri: The store URI string
-    :param stores: The SessionStores object that the new stores will be added to
+    :param stores_uri: The store URI.
+    :param stores: The SessionStores object that the new stores will be added to.
```
sei-aderr: `UnifiedStore` type

test/session/conftest.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def create_test_session_stores(
+    store_type: StoreType,
+    tmp_path,
+    patched_setup_stores,
+    catalog_uris: dict[str, StoreURI] = {},
+) -> UnifiedStore:
+    """Creates appropriate test session stores."""
+    # We can't test setup_stores with a REMOTE store, since this would require setting up the session stores,
```
sei-aderr: This is now `UnifiedStore` not `SessionStore`

test/store/user/test_underlying.py
```diff
@@ -87,40 +79,50 @@ def get_test_permissions() -> List[Permission]:
     return [p1, p2]
 
 
+def get_internal_store_session(
```
sei-aderr, line 82: How is this working separately from `create_test_session_stores`? That one says we can't differentiate between local and remote stores but this looks to work around it somehow?

test/store/user/test_underlying.py
```diff
@@ -87,40 +79,50 @@ def get_test_permissions() -> List[Permission]:
...
+def test_user(store_type: StoreType, create_test_user_store) -> None:
     """An artifact store supports user operations."""
-    store: UserStore = request.getfixturevalue(store_fixture_name)
+    store: UserStore = create_test_user_store(store_type)
 
     test_user = get_test_user()
     email2 = "email2@server.com"
     name2 = "new name"
 
+    # if store_type
```
sei-aderr: Dangling comment

test/store/user/test_underlying.py
```diff
@@ -177,14 +184,21 @@ def test_user_group_change(
...
         # Ensure user has updated group info.
+        found_group = None
         read_user = user_store.user_mapper.read(test_user.username)
-        assert read_user.groups[0] == updated_group
+        for group in read_user.groups:
+            if group.name == updated_group.name:
+                found_group = group
+                break
 
+        # If we got here, the updated group was not even found.
```
sei-aderr: Wouldn't this always be reached?

sei-aderr (changes requested):
Looks good overall, lots of changes so had lots of questions.

mlte/store/base.py
sei-secheverria: That is a good point, but it is a practical method, and I'd argue that it could be useful to people using MLTE, which is what the tests do. I'd rather leave it as it makes sense from a user's perspective.

test/backend/api/endpoints/test_catalog_entry.py
sei-secheverria: Will improve

test/store/utils.py
sei-secheverria: It's a bug or inconsistency, will check

mlte/artifact/artifact.py
sei-secheverria: Will fix

test/store/artifact/conftest.py
sei-secheverria: Yes, that is a left over from when the artifact store was the only one. Will fix.

test/store/artifact/conftest.py
sei-secheverria: The user one is the one that, intentionally, needed a cached version, to avoid having to calculate the hashed passwords each time, since that is so slow. I am not sure if the catalog one needs it or not, there may be a similar reason that I can't remember right now.

test/store/user/conftest.py
sei-secheverria: leftover

mlte/backend/api/endpoints/user.py
sei-secheverria: The mapper returns a User, and this function ends up returning a User, so there is no need for explicit type hinting. Plus the new_user is only used to get its id to re-read the user. Before I think it was used for more checks here, and the types were different. So nothing to do that I can see.

mlte/backend/api/endpoints/user.py
sei-secheverria: Maybe the function name is not very clear. What this is doing is removing any new groups from the received user. It is a weird permission we have that does not have a standard way to be used through our regular permission checks. What we are saying is that only an Admin user can change a user groups. So if we get a user and the current user is not admin, we remove any new groups that may have been added.

mlte/store/catalog/underlying/http.py
sei-secheverria: Yeah, best I could do for now. This is not really new, I fixed a bug there and added the explanation, but it has been there for a while. Maybe one day we'll make it prettier, but it will require major refactoring of the base store framework.

mlte/store/user/policy/policy_store.py
sei-secheverria: yep

mlte/store/user/policy/policy_store.py
sei-secheverria: Will do

mlte/store/user/underlying/rdbs/store.py
sei-secheverria: For HTTP, it is a different type of store. Unlike all the other ones, it assumes there is another type of store remotely, in another backend that was intiialized properly. Thus, any initialization was done there by that backend. If an HTTP store sets up default data, it will duplicate what the remote store already did. It does not make sense for it.

Now, it does make sense to handle this in the HTTP store itself, true, it makes no sense to do it as it is now and must be a leftover. Will fix.

FS does call super (line 46).

test/artifact/test_artifact.py
sei-secheverria: If I understand correctly, that is only checked by the crossvalidators, if set, which are only set if the UnifiedStore is used. For these tests, the artifact_store_with_context fixture is used, which just directly creates an artifact memory store without going through the UnifiedStore. Thus, validators are not set.

I guess if we wanted to fix this, we would  have to change artifact_store_with_context to use the same unified store methods, using the patched setup_stores. But I doubt that is the only place where that would be needed, and I am not sure we need to check that here anyway.

Now, thinking about it, should the validators be an argument of the factory store methods, rather than a separate method that one has to remember to call?

test/artifact/test_artifact.py
sei-secheverria: It is setting a pre-created store into the UnifiedStore in the Session object. The session methods assume you pass a URL and create the stores, in these cases the stores were already created.  Maybe this could be made more efficient using the new features, changing how artifact_store_with_context is used, but I'd leave that for later. I haven't really touched it.

test/store/user/test_underlying.py
sei-secheverria: This is a workaround since the User store tests need to do stuff directly on the store, which can't be done through the HTTP store in the case we are testing that type of store. 

The create_test_session_stores doesn't say you can't differentiate, it just states that for the remote HTTP store, you don't can't create two stores due to the way the "state" object is a global thing in the backend/api.

test/store/user/test_underlying.py
sei-secheverria: Outdated comment

mlte/store/base.py
sei-aderr: That works, if we want it to be available user facing then we should document it somewhere

mlte/backend/api/endpoints/user.py
sei-aderr: Function name was confusing, but I was also thinking about the logic wrong. Makes sense looking at it again

mlte/store/user/policy/policy_store.py
sei-aderr: `PolicyStoreService` works for me, could you adjust the file name to match?

test/artifact/test_artifact.py
sei-aderr: That is true, yeah they would not be setup in that case.

I am not sure having them be part of the factory store methods would work. I think we would end up with circular imports because each factory would then depend on the store types of the cross validators it uses. So `unified_store` would then circular import. There may be a way to adjust that, but already ran into issues with it and that is why we have the `CompositeValidator` class to act as an intermediary that holds the `CrossValidators` without their implementations that are in separate files

sei-aderr: If it did work, I think that sounds like a good idea? Another issue that I ran into making the validators was that it becomes very order dependent so the stores that are needed for validation are created before the ones that use them, I don't think that would present any issues with this change as it would still just have to be in the same order.

I guess the issue this then presents though is any time that you want to one off create a store that has validators, you would need to make the other associated stores. Although if the validators were a param, we could probably just have them be optional so that it doesn't totally rely on having those other stores if we don't want them for a simple test.

So yeah, could give it a shot and see if it works without much issue. Then we could have all of the creation of the stores stay within the factories so you don't have to do extra steps around it to get a fully filled out one

mlte/artifact/artifact.py
sei-aderr: Looks like this one may have been missed, did you have any thoughts on this?

mlte/store/base.py
sei-secheverria: In theory yes, but we are not really documenting most of that (we are not documenting the from_string method that has already been there for a while). This is roughly at the same level. If we want to better document all these methods, we could. Do note that in the read the docs, we do have an API reference (https://mlte.readthedocs.io/en/latest/reference/store/base/), which is probably enough for now.

test/artifact/test_artifact.py
sei-secheverria: This may be worth it, but I'd spin it to a separate ticket. Doesn't seem critical to this PR and it could take a bit, and I am still not totally convinced this is the way to go and may want to rethink it for a bit.

mlte/artifact/artifact.py
sei-secheverria: Yeah, read it but forgot to comment. My initial thoughts is that I wanted to leave that open for compatibility, in case there were cases I had not checked where it made sense. My other thoughts were that some people may not care about users if they are doing quick and dirty local things. It is true that we could cover those cases by admin. But also, do note that the current credentials are optional. Users may not set up the env vars or call the methods to set up users or password. If they don't, the user is actually None. We could force it to be admin, but I am not sure that is right and may end up being a security issue. If we do remove the possiblity of having no user tracked in the stores, we would have to force people to set the credentials, and I am not totally sure we want to do that without some default user (which admin may or may not be appropiate for.).

test/artifact/test_artifact.py
sei-aderr: Works for me

mlte/store/user/policy/policy_store_service.py
```diff
@@ -0,0 +1,77 @@
...
+
+from __future__ import annotations
+
+import mlte.store.error as errors
+from mlte.store.user.mappers import GroupMapper, PermissionMapper
+from mlte.store.user.policy import Policy
+
+
+class PolicyStore:
+    """Handles access to stored policies. It is not a store in the same sense as the rest, does not derive from Store."""
```
sei-secheverria: Ok, pushed changed filename

test/artifact/test_artifact.py
sei-secheverria: If you could create the issue, I'd appreciate it

sei-aderr: Added #865

sei-aderr (approved):

Issue #872, opened 2026-04-22
Try out using LLMs to generate TestSpecification


sei-secheverria:
Use LLMs to see if TestSpecification can be automatically generated. Sources for this:

 - The model's NegotiationCard, specifically the QA requirements sections for the most part.
 - Knowledge of the Validators and Measurements present in the MLTE codebase.
 - Potentially some of MLTE's docs.

sei-lzhan:
```
Generate valid Python source code from the negotiation card data below.



Your response must be directly copy-paste ready Python.



Hard output rules:

- The first line of your response must be exactly: `spec = TestSuite(`

- The last line of your response must be exactly: `spec.save(parents=True, force=True)`

- Return no text before the Python.

- Return no text after the Python.

- Do not return markdown.

- Do not return code fences.

- Do not return explanations.

- Do not return summaries.

- Do not return reasoning.

- Do not return any conversational text.



Generate a complete test suite:

- `spec = TestSuite(test_cases=[ ... ])`

- `spec.save(parents=True, force=True)`



Each generated `TestCase(...)` must include:

- `identifier`

- `goal`

- `quality_scenarios`

- `validator`



Include `measurement` only when the scenario clearly fits:

- a built-in measurement

- a known evidence type

- a standard external function already implied by the prompt



Include `note` when the scenario is better written as:

- a test plan

- a comparison plan

- a review plan

- a statistical check

- a pass/fail rule

- a data setup plan



When `note` is used, prefer this structure:

- `Data%DataSource: ...`

- `Measurement&Condition: ...`



Measurement API context:

- A measurement object is used to produce the result for a test case.

- `ExternalMeasurement` wraps an external Python function and converts its result into a chosen evidence type.

- `ImportMeasurement` imports external JSON/dict-like results as a general evidence type.

- Prefer validators from the evidence type or the measurement output type when available.

- Prefer a concrete measurement when the scenario clearly fits one.

- If no built-in measurement clearly fits, use `ExternalMeasurement(...)`.

- For APIs with uncertainty, add a comment to signal reviewer for manual inspection.



Allowed existing symbols:

- `TestSuite`

- `TestCase`

- `ExternalMeasurement`

- `ImportMeasurement`

- `Real`

- `ConfusionMatrix`

- `Image`

- `LocalObjectSize`

- `LocalProcessMemoryUtilization`

- `LocalProcessCPUUtilization`

- `Units`

- `validators`

- `model`



Core rules:

[...]

PR #878, opened 2026-05-14
Implemented test set to verify general LLM TestSuite format

### LINKED ISSUE #875: Create code to validate output from LLM
Create a representative test set to validate LLM generated Test Suit.

The test set should cover different types of negotiation cards and define the baseline for acceptable output that is ready for human-in-the-lead review and implementation.

sei-lzhan:
Things included in this PR:

- Verify generated output has the correct MLTE test suite structure
- Make sure the output contains only allowed MLTE symbols
- Make sure each test case has quality_scenarios field and the format is correct (a list with the correct *-qas format)
- Ensure necessary comments on quality attribute and customized validators are properly documented in the LLM output

Steps to test:
- Generate LLM output by using the prompt template in [this issue](https://github.com/mlte-team/mlte/issues/872)
- Paste result in test/generated_llm_outputs/llm_generated_test_suite.py
- To save time and not run the full test, you can run this single test by running `python -m pytest test/llm/test_llm_generated_test_suite.py -q`

Questions to @sebastian-echeverria 
Pasting result in a specific file seems a bit hardcoded. But given that we need to obtain the result from LLM UI, this seems like the most straight forward solution for now. Potentially, we can set up a full API call pipeline and obtain the output to local file directly but that seems to be a bit out-of-scope for this task. Thoughts on alternative/better approach?

### DIFF
--- a/test/llm/test_llm_generated_test_suite.py
+++ b/test/llm/test_llm_generated_test_suite.py
@@ -1,8 +1,9 @@
 import ast
 import os
 import re
-import pytest
 from pathlib import Path
+
+import pytest
 
 GENERATED_OUTPUT_DIR = "test/generated_llm_outputs"
 
--- a/test/llm/test_llm_generated_test_suite.py
+++ b/test/llm/test_llm_generated_test_suite.py
@@ -451,7 +451,7 @@ def get_symbol_root(syntax_node: ast.AST) -> str | None:
 
 
 def get_attribute_root(attribute_node: ast.Attribute) -> str | None:
-    current_node = attribute_node
+    current_node: ast.expr = attribute_node
 
     while isinstance(current_node, ast.Attribute):
         current_node = current_node.value
--- a/test/llm/test_llm_generated_test_suite.py
+++ b/test/llm/test_llm_generated_test_suite.py
@@ -1,0 +1,473 @@
+import ast
+import os
+import re
+from pathlib import Path
+
+GENERATED_OUTPUT_DIR = "test/generated_llm_outputs"
+
+REQUIRED_TESTCASE_FIELDS = {
+    "identifier",
+    "goal",
+    "quality_scenarios",
+    "validator",
+}
+
+OPTIONAL_TESTCASE_FIELDS = {
+    "measurement",
+    "note",
+}
+
+ALLOWED_TESTCASE_FIELDS = REQUIRED_TESTCASE_FIELDS | OPTIONAL_TESTCASE_FIELDS
+
+QAS_ID_PATTERN = re.compile(r"^[A-Za-z0-9_.-]+-qas_\d{3}$")
+
+BASE_ALLOWED_SYMBOLS = {
+    "spec",
+    "TestSuite",
+    "TestCase",
+    "ExternalMeasurement",
+    "ImportMeasurement",
+    "Real",
+    "Integer",
+    "String",
+    "Array",
+    "Opaque",
+    "Image",
+    "ConfusionMatrix",
+    "LocalObjectSize",
+    "LocalProcessMemoryUtilization",
+    "LocalProcessCPUUtilization",
+    "NvidiaGPUMemoryUtilization",
+    "NvidiaGPUPowerUtilization",
+    "Units",
+    "validators",
+    "model",
+    "ProcessMeasurement",
+    "ProcessMeasurementGroup",
+    "CommonStatistics",
+    "CPUStatistics",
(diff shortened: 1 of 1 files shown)


test/llm/test_llm_generated_test_suite.py
```diff
@@ -0,0 +1,485 @@
+import ast
+import os
+import re
+from pathlib import Path
+
+import pytest
+
+GENERATED_OUTPUT_DIR = "test/generated_llm_outputs"
```
sei-secheverria, line 8: I don't think we should use this folder. "test" is for code that tests the platform; if this is a temporary file with a generated TestSuite, it should be in a temporary folder. Or if we want to have it in a way that we can easily access it, it should be in some dedicated folder that is included in .gitignore. Maybe inside test/temp, then have the "generated..." folder inside, and add test/temp to gitignore?

test/llm/test_llm_generated_test_suite.py
```diff
@@ -0,0 +1,485 @@
...
+    "Units",
+    "validators",
+    "model",
+    "ProcessMeasurement",
+    "ProcessMeasurementGroup",
+    "CommonStatistics",
+    "CPUStatistics",
+    "MemoryStatistics",
+    "NvidiaGPUMemoryStatistics",
+    "NvidiaGPUPowerStatistics",
```
sei-secheverria, line 53: Even though this works for an initial test, it will never scale as we would have to maintain this somehow. Not sure how to best handle this though.

test/llm/test_llm_generated_test_suite.py
```diff
@@ -0,0 +1,485 @@
...
+def assert_generated_output_has_required_format(
+    generated_source_code: str,
+) -> None:
+    generated_source_lines = generated_source_code.splitlines()
+
+    if not generated_source_lines:
+        raise AssertionError("Generated output is empty")
+
+    first_source_line = generated_source_lines[0]
+    last_source_line = generated_source_lines[-1]
```
sei-secheverria, line 113: There should be an easier way to test this out, or at least to simplify this so that we take advantage of existing support to have to test less things here. Some ideas:
 - TestSuites are an object, but as al artifacts in MLTE, they can be serialized into JSON. There are many ways we could take advantage of this:
    - If instead of having to check the whole TestSuite, we tried to convert this TestSuite to JSON, the conversion process itself takes care of checking it against the TestSuite JSON model schema, so it would flag a lot of the errors (possible not some of the validators/measurements, would have to check)
    -  Another interesting option, but I am not sure if it would be easier or not, would be to modify the prompt so that the LLM generates the JSON version of the TestSuite, not the code version. This has some pros and cons. A pro is that it should be much easier to test, as most of it (though maybe not all, things like validators) should be easy to test by just loading a TestSuite object from the JSON file. It also would allow us later to just add a TestSuite as a JSON file directly to the store, without having people to copy/paste the results of the JSON code to use it. This is sometimes useful if you want to define the TestSuite in a JupyterNotebook, for example, so we should discuss this.

sei-secheverria: The main reason I am thinking about that is that this test class seems fragile and will be hard to maintain. I'd look into the pros/cons of the JSON based option before moving forward with this. If we do decide to do this not in JSON format, we can talk about ways of making it more robust.

test/llm/test_llm_generated_test_suite.py
sei-secheverria, line 1: One thing that is missing here is a more abstract thing. We do not only need to have functions to test, we need a test set in terms of NegotiationCard / TestSuite pairs that are appropriate test cases for this test functions to execute. For this, please initially think about what cases these should cover and we can discuss them, and then we can create them as JSON files. One thing is testing to check that the output is valid, but another is to check that it is appropriate. You can split that into a separate issue to better track it, though.

sei-secheverria (changes requested):
Looks nice, but fragile. Some ideas to improve this below, as well as a potentially missing separate issue about defining the test data.

Regarding this question: "Pasting result in a specific file seems a bit hardcoded. But given that we need to obtain the result from LLM UI, this seems like the most straight forward solution for now. Potentially, we can set up a full API call pipeline and obtain the output to local file directly but that seems to be a bit out-of-scope for this task. Thoughts on alternative/better approach?"

Yes, we want to be able to use this in an integrated way, which means we want to set up an API call pipeline. I wouldn't say it is out of scope, but it is probably better to create another issue to look into that.

PR #896, opened 2026-07-09
Feature/store import

### LINKED ISSUE #845: Basic Import Functionality
Corresponding import to export in #843. 

Care should be taken when importing. Some logic should be used to handle resources that already exist in the store where things are being imported to, including custom lists, artifacts, users, etc. One option would be to make the import atomic, and to just fail if any resource is already there. Another would be to overwrite everything (and both options could be combined with an overwrite flag). Another would be to import everything but avoid overwriting dependencies that are already there.

sei-aderr:
Resolves #845


mlte/store/user/policy/user_policy.py
```diff
@@ -5,12 +5,18 @@
...
+    User,
+    UserWithPassword,
+)
 
 
 def set_default_user_policies(
-    user: UserWithPassword, policy_store: PolicyStoreService
-) -> UserWithPassword:
+    user: Union[User, UserWithPassword], policy_store: PolicyStoreService
+) -> Union[User, UserWithPassword]:
```
sei-secheverria, line 19: This is good, but just as a coincidence, we are moving away from the `Union[A, B]` and into the `A | B` format, which is strongly recommended in general, and more so by uv/ruff, so it is probably a good idea to start structuring that way already.

mlte/store/user/underlying/rdbs/store.py
```diff
@@ -122,7 +122,9 @@ def __init__(
         self.policy_store = policy_store
         """Policy store abstraection"""
 
-    def create(self, user: UserWithPassword, context: Any = None) -> User:
+    def create(
+        self, user: Union[User, UserWithPassword], context: Any = None
```
sei-secheverria: Same comment, switch to `User | UserWithPassword` to future-proof

mlte/store/user/underlying/fs.py
```diff
@@ -111,18 +111,23 @@ def __init__(
         self.policy_store = policy_store
         """Policy store abstraction."""
 
-    def create(self, user: UserWithPassword, context: Any = None) -> User:
+    def create(
+        self, user: Union[User, UserWithPassword], context: Any = None
```
sei-secheverria: Same Union comment

mlte/store/user/underlying/http.py
```diff
@@ -106,7 +106,9 @@ def __init__(self, storage: HttpResourceStorage) -> None:
         self.storage = storage
         """The HTTP storage access."""
 
-    def create(self, new_user: UserWithPassword, context: Any = None) -> User:
+    def create(
+        self, new_user: Union[User, UserWithPassword], context: Any = None
```
sei-secheverria: Same about Union

mlte/store/user/underlying/memory.py
```diff
@@ -115,21 +115,26 @@ def __init__(
         self.policy_store = policy_store
         """Policy store abstraection"""
 
-    def create(self, user: UserWithPassword, context: Any = None) -> User:
+    def create(
+        self, user: Union[User, UserWithPassword], context: Any = None
```
sei-secheverria: Same about Union

mlte/store/user/mappers.py
```diff
@@ -11,7 +11,9 @@
 class UserMapper(ResourceMapper):
     """An interface for mapping CRUD actions to store users."""
 
-    def create(self, new_user: UserWithPassword, context: Any = None) -> User:
+    def create(
+        self, new_user: Union[User, UserWithPassword], context: Any = None
```
sei-secheverria: Same

mlte/store/import_export/import_store.py
```diff
@@ -0,0 +1,275 @@
...
+def import_store(
+    import_data: dict[str, Any],
+    artifact_store: ArtifactStore,
+    custom_list_store: CustomListStore,
+    user_store: UserStore,
+    catalog_stores: CatalogStoreGroup,
+    force: bool = False,
+) -> None:
+    """
+    Import MLTE store objects from json dict.
```
sei-secheverria, line 35: This really needs to explain what each argument is, especially the import_data one.

mlte/store/import_export/import_store.py
```diff
@@ -0,0 +1,275 @@
...
+
+
+def _import_users(
+    user_data: dict[str, Any], user_store: UserStore, force: bool = False
+) -> None:
+    """Import user data into store."""
+    with ManagedUserSession(user_store.session()) as user_store_session:
+        user_name_list = user_store_session.user_mapper.list()
+
+        # how do we go from a "User" that has the hashed password that we are importing, to "UserWithPassword" that expects plain text passowrd?
```
sei-secheverria: I think this has already been fixed?

sei-secheverria (changes requested):
I think it looks good. There are a couple of minor typehint and comment adjustments I recommend below, very simple.

sei-secheverria (approved):

PR #897, opened 2026-07-09
Feature/uv migration

### LINKED ISSUE #693: Move from Poetry to uv
Consider moving from poetry to uv as dependency management system.

### LINKED ISSUE #695: Find easier way to get output type of validator for TestSuite creation
Call to value()/get_output_type().validator_method is too wordy. It would be nicer to call them as static objects of base measurement class. However, this doesn't work as validators are on Evidence types, not Measurement class. See if there is a way to improve this.

sei-secheverria:
Migrates to uv/ruff, and makes a couple more changes.
 - Fixes #693 by moving from poetry to uv, and updating all commands, docs, etc to use it instead. This also simplifies some steps and pyenv can be removed from instructions.
 - Along with change to ruff, several lint/formatting improvements recommended by its default config were implemented.
 - Lint/formatting was also applied now to all code in demo/ folder (before it was only affecting some of this code)
 - Some Makefile commands were optimized or ordered.
 - pyproject.toml was updated to a more generic structure, and dev groups were subdivided for clarity
 - Dev dependencies were updated, and some runtime ones were as well
 - Building of Docker images for dist and ops was added to CI configuration.
 - Also fixes #695 to call the method with the output type just output() (instead of get_output_type()) for a more readable experience when using it in the validator attribute of a TestCase.
 - Bug in one of the pynvml utils tests was fixed (mock was hiding proper exception).
 - Some internal methods and attributes were renamed to avoid name clashes with Python keywords and weird type checking workarounds.


Makefile
```diff
@@ -1,141 +1,164 @@
...
-# Lint all source code
-.PHONY: lint
-lint:
-	poetry run flake8 mlte/
-	poetry run flake8 test/
-	poetry run flake8 tools/
+	uv run ruff format --check .
 
-# Typecheck all source code
+# Typecheck all source code (except for demo).
```
sei-aderr, line 77: Why is this one excluded? I see that the other QA now look at the demo folder

sei-aderr: And I see that it is a separate command `demo-typecheck` a bit further down

pyproject.toml
```diff
@@ -1,45 +1,56 @@
...
+    "fastapi >=0.115.6",
+]
+backend = [
+    "mlte[web]",
+    "python-jose >=3.3.0",
+    "python-multipart >=0.0.20",
 ]
-gpu = [
-  "nvidia-ml-py >= 13.580.82",
+frontend = [
```
sei-aderr, line 50: Why did this get made a list of 1 and rename frontend to web?

docker/Dockerfile.python
```diff
@@ -20,16 +20,16 @@ RUN pipx ensurepath
 # Ensure we are not buffering output for clarity.
 ENV PYTHONUNBUFFERED=1
 
-# Setup poetry to handle dependencies.
-RUN pipx install poetry==2.0.1
+# Setup uv to handle dependencies.
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
```
sei-aderr: Do we want to pin this to a specific version instead of latest to avoid any unintentional breaking updates to uv?

Makefile
sei-secheverria: The demo stuff can be installed separtely, and it has a lot of requirements. The only reason it was separated was to better group it, so if someone doesn't want to install the demo requirements because they won't be working on the demos, they won't have to.

sei-secheverria: I don't recall why mypy was treated differently. May be something to do with the extra type libraries that are needed for mypy. There was some incompatibility and this made things cleaner.

pyproject.toml
sei-secheverria: The point is not really that it was renamed. Two separate groups were created, one for backend and one for frontend. The idea is to allow people that want to install only the frontend to only have to install depenendecies for that, and those that only want to install the backend to install only those dependencies, in case someone wants one but not the other. The "web" is a common group for the dependdencies shared between frontned and backend.

docker/Dockerfile.python
sei-secheverria: Good question, we probably do. I'll take a look.

sei-secheverria: Ok, updated and pinned

Makefile
sei-aderr: Running mypy on the demo folders wouldn't require the demo libraries though, would it? If there was some issue with it, this works just seems odd at first glance

sei-secheverria: To be honest, I don't recall the exact issue. But running mypy not only requires the dependencies, but also in some cases desparate type dependencies as well. Mypy needs to check types, and those types are defined sometimes in the dependencies themselves, sometimes in additional libraries.

sei-aderr (approved):

PR #907, opened 2026-08-11
Artifact Overwrite Unit Test Bug Fix

### LINKED ISSUE #736: Bug in Artifact unit tests
There is currently a bug in the artifact unit tests that only sometimes shows up in the CI runs. The bug is in test_artifact_overwrite (at least with [http_store-report). The error is a ErrorAlreadyExists that is expected but does not always show up. An example failed run can be seen here: https://github.com/mlte-team/mlte/actions/runs/18234729778/job/51926134193

For some reason, sometimes the check for existence is not working properly. As this only sometimes happens, it is hard to reproduce, and it sounds like the is a race condition somewhere. Usually when re-running the job, the error goes away.

It is not clear exactly in which case it happens, but it is possible it only happens with the HTTP store. There may be a delay in the API getting the HTTP request that creates the artifact, and by the time the next request is sent to check if the artifact is there, it may not have been created yet. However, this is just a theory.

sei-secheverria:
Fixes #736, and removes some testing warnings. More specifically:

 - Issue #736 was due specifically to how Report ids work. The problem always happened in the overwrite test for a report. The problem is that, by definition, reports can never be overwritten, as, whenever they are saved, their id is updated to add a timestamp. Thus, the test was not valid for reports. But since the timestamp has seconds-level granularity, it could happen that, between writing and overwriting a report, the timestamp could change by 1 second or not, depending on the timing of the test. In the first case (same timestamp generated as still running in the same second) the test would pass, in the second case (new timestamp generated since a second passed) the test would fail as the new report would have a new id and the check to see if it existed failed. Fixed by simply skipping that test for Reports, since it does not make sense for that artifact type.
 - Also fixed a related bug, in that the hook adding the timestamp to a report id could append more timestamps after previous ones. While a case not in a regular workflow, it could happen. Now, it removes any potential timestamp from a report id before appending a new one.
 - Also fixed warning from newer pytest versions of not using generators as parametrized values (and converting them to lists instead).

### DIFF
--- a/mlte/report/artifact.py
+++ b/mlte/report/artifact.py
@@ -3,8 +3,8 @@
 from __future__ import annotations
 
 import datetime
-import typing
 import re
+import typing
 
 from mlte.artifact.artifact import Artifact
 from mlte.artifact.model import ArtifactModel
@@ -142,7 +142,7 @@ class Report(Artifact):
         :param store: The store in which to save the artifact
         """
         # Ensure that, if the id already had a timestamp, it is removed first.
-        timestamp_suffix = datetime.datetime.now().strftime('%Y%m%d-%H%M%S')
+        timestamp_suffix = datetime.datetime.now().strftime("%Y%m%d-%H%M%S")
         cleaned_id = re.sub(r"-\d{8}-\d{6}$", "", self.identifier)
         self.identifier = f"{cleaned_id}-{timestamp_suffix}"
 
--- a/mlte/report/artifact.py
+++ b/mlte/report/artifact.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import datetime
 import typing
+import re
 
 from mlte.artifact.artifact import Artifact
 from mlte.artifact.model import ArtifactModel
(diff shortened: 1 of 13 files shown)


mlte/report/artifact.py
```diff
@@ -139,9 +140,11 @@ def pre_save_hook(self, context: Context, store: ArtifactStore) -> None:
         Override Artifact.pre_save_hook(). Assigns time-stamped id to report, to ensure all have different ids.
         :param context: The context in which to save the artifact
         :param store: The store in which to save the artifact
-        :raises RuntimeError: On broken invariant
         """
-        self.identifier = f"{self.identifier}-{datetime.datetime.now().strftime('%Y%m%d-%H%M%S')}"
+        # Ensure that, if the id already had a timestamp, it is removed first.
```
sei-aderr, line 144: So when would this get called twice in such a way that the id would already have a timestamp and then get called again to add another?

sei-secheverria: If someone writes a report, then reads it, tries to modify it and write it again. Not a standard flow, but just covering that case.

sei-aderr: Got it, thought it was some weird case that it could happen under normal use. Good catch on the rest of the fixes! Hopefully this fully solves it

sei-aderr (approved):

PR #908, opened 2026-08-11
Feature/card UI improvements

### LINKED ISSUE #829: Improve Information about Negotiation Card Context in Detailed View
Currently, the Negotiation Card detailed view, where it can be edited, is seen like this:

Image

There are a couple of usability problems with this view:
- From this view, there is no way to figure out for what model/version the negotiation card we are looking at is. In the main artifact store, this is provided by the dropdowns, but here this is lost. It would be nice to add a way to show that information up top.
- The main title, after "Negotiation Card", is "card.default". That is so ugly. Most likely, people are going to be using the default ids for most artifacts, and that will be card.default for the card. We should make this more user friendly. Putting "card.default" as a main title does not help the user. There are different ways to improve this, but one way would be to showcase more the missing model/version info, and make the id less prominent (such as, id: card.default, but not in title font).
- It is very likely that this also applies to  Reports in the same way, though the id is slightly more useful in that case as it shows the date, but it still ugly. This also applies to other artifacts views, such as evidence, which are less critical bit could also benefit from this change.

### LINKED ISSUE #861: Show currently selected page in menu
The frontend does not currently highlight the page/section it is in in its menu. Add this.

sei-aderr:
Resolves #829 resolves #861 

Resolves two tickets, removes outdated TEC import that isn't used, makes some improvements to the base-layout


sei-secheverria (changes requested):
Running in docker is failing for some reason, both in the CI and locally for me. The lock file for the frontend is out of date maybe?

sei-secheverria:
The lock issue seems to be fixed, but there is still something wrong as CI is reporting an issue with a svg file not being found during build time.

sei-aderr:
@sebastian-echeverria 

Have you seen these issues that are happening with readthedocs? I updated twine because it looked like we may be getting a version of hatchling that isn't pinned, and the version of twine we had didn't support the new 2.5 metadata version it was being given. However, now it looks like uv is having issues on readthedocs

sei-secheverria:
> @sebastian-echeverria
> 
> Have you seen these issues that are happening with readthedocs? I updated twine because it looked like we may be getting a version of hatchling that isn't pinned, and the version of twine we had didn't support the new 2.5 metadata version it was being given. However, now it looks like uv is having issues on readthedocs

I saw the twine issue and I think it is now fixed in master, since I updated it in a recent PR. But I have not seen the readthedocs issue, what is the problem there?

sei-aderr:
It just looks like the uv install is failing in read the docs for some reason, it says it is already installed and then says it can't find any versions

https://app.readthedocs.org/projects/mlte/builds/34038523/

uv looks to be installed by default on readthedocs so I tried to update it the new format that it expects, but that didn't seem to work either, so I reverted that. I think I may try and get this new version to work though as it seems to be the intended way with uv

https://docs.readthedocs.com/platform/stable/config-file/v2.html#python-install

I haven't used readthedocs much so was just wondering if you had seen anything similar, or had any immediate ideas

sei-secheverria:
> It just looks like the uv install is failing in read the docs for some reason, it says it is already installed and then says it can't find any versions
> 
> https://app.readthedocs.org/projects/mlte/builds/34038523/
> 
> uv looks to be installed by default on readthedocs so I tried to update it the new format that it expects, but that didn't seem to work either, so I reverted that. I think I may try and get this new version to work though as it seems to be the intended way with uv
> 
> https://docs.readthedocs.com/platform/stable/config-file/v2.html#python-install
> 
> I haven't used readthedocs much so was just wondering if you had seen anything similar, or had any immediate ideas

I don't know what the issue is now, but I while ago, when updating to uv, I had to move from the mkdocs library to properdocs, since mkdocs is no longer maintiained as we are using it. Since readthedocs had built in support for mkdocs and not properdocs, I had to do some workarounds in the installation for it to work properly, couldnt use the simpler jobs and built-in things. But I am not sure why it would not work now, what would trigger that. Some dependency?

sei-secheverria:
> > It just looks like the uv install is failing in read the docs for some reason, it says it is already installed and then says it can't find any versions
> > https://app.readthedocs.org/projects/mlte/builds/34038523/
> > uv looks to be installed by default on readthedocs so I tried to update it the new format that it expects, but that didn't seem to work either, so I reverted that. I think I may try and get this new version to work though as it seems to be the intended way with uv
> > https://docs.readthedocs.com/platform/stable/config-file/v2.html#python-install
> > I haven't used readthedocs much so was just wondering if you had seen anything similar, or had any immediate ideas
> 
> I don't know what the issue is now, but I while ago, when updating to uv, I had to move from the mkdocs library to properdocs, since mkdocs is no longer maintiained as we are using it. Since readthedocs had built in support for mkdocs and not properdocs, I had to do some workarounds in the installation for it to work properly, couldnt use the simpler jobs and built-in things. But I am not sure why it would not work now, what would trigger that. Some dependency?

Still not sure why this broke, but if the issue is the line "asdf global uv latest", we could start pinning the uv version, which we pinned anyway in our Dockerfiles to 0.12 . If that doesn't work, it could be an issue on their side.

sei-secheverria (approved):

PR #912, opened 2026-08-25
Feature/handle errors

### LINKED ISSUE #75: Measurement Error Handling
We need to report errors in a uniform manner within `Measurement`s. In order to do this, we need to implement an `Exception` hierarchy in a location that is accessible across `Measurement`s.

In addition, or alternatively, we could log errors, and do other related handling. But mostly we should define a unified way of handling this that matches our current use cases.

sei-secheverria:
Resolves #75 by adding an optional parameter to Measurements to generate an Unavailable type of Evidence when an error occurs in a Measurement, instead of throwing an exception and stopping the process. This is off by default, but can be easily turned on by a Measurement attribute. TestCases' measure() method, however, does have it enabled by default, as that method was created to be used when running all test cases in an automated way, where storing details of a measurement error is more important that stopping the whole process.

Also fixes minor doc issue with command to publish packages to pypi using uv.


sei-aderr:
I think this generally looks reasonable, but how does the process continue after an unavailable evidence is created? So if you run a test suite and get an unavailable it will "silently" fail to and create that instead of erroring out, so if you then create a report will it just include that unavailable evidence in the report? With just this implementation it also seems like the unavailable would show up in the frontend under the evidence list like all the others

sei-secheverria:
> I think this generally looks reasonable, but how does the process continue after an unavailable evidence is created? So if you run a test suite and get an unavailable it will "silently" fail to and create that instead of erroring out, so if you then create a report will it just include that unavailable evidence in the report? With just this implementation it also seems like the unavailable would show up in the frontend under the evidence list like all the others

Correct, that is the idea. It is equivalent to how pytest will run all test cases, and will complain if something weird happened to one. The idea is that if you want the process to stop as you run it, by default it will, as it does now. If you don't, the process will continue, store information about the error in the Unavailable evidence, and show in the result as a Failure (as Unavailable evidences are automatically treated as a Failure), which a descriptive message indicating why it failed. This allow people that want to contiunally or in an auomated way to run all test cases to look at the results later and use the info in the report + the evidence to figure out why a run failed in a specific measurement, without having stopped running other cases.

sei-aderr (approved):

PR #914, opened 2026-09-01
889 build representative validation test set

### LINKED ISSUE #889: Build Representative Validation Test Set
We should manually create a set of negotiation cards to cover different kinds of measurements that we aim to test. We can have a combination of one measurement and multiple measurements in different negotiation card examples. Finally, we need to craft input & output that can be used to validate each negotiation card.

sei-lzhan:
Resolves #889


sei-secheverria (changes requested):
There are a couple of major changes that are needed before I can review this PR:
 - First, this branch is from a very old version of main, and we've changed a lot of the qa between then and now. Please either merge main into your branch or rebase it so it can start with a newer base.
 - Second, checks are failing due to not fixing qa issues. But you should first update to the current version of main before figuring that out.

sei-lzhan:
Sorry about the delay! Both issues have been addressed, ready for review

sei-secheverria (changes requested):
The data sets look great. The only change I would make is to try to test out more built-in validators. We are currently testing the resource-specific ones, which is great, plus some of the Real validator methods. However, there are some other built-in validator methods that are different, such as the String related ones, or the Array and Image ones. I would maybe add another example (and/or modify some of the existing ones) to try to cover those types of Validators as well.

PR #917, opened 2026-09-02
Feature/dataset import

### LINKED ISSUE #915: Add way to import dataset information from file
Collaborator has dataset with 100 different fields in it, this is incredibly painful to input manually. Look into ways that this could be automatically imported into a negotiation card (through the frontend).

Considerations
- How to structure data coming in, how to validate it and make it a reusable structure we expect
- How to handle the import, do we do it all in the frontend and then auto-save? We probably want validation to happen in the backend and then send back in some way, but that would probably require a page reload

sei-aderr:
Resolves #915 

Some minor frontend updates, then adds csv importing for the data schema.

### DIFF
--- a/docker/Dockerfile.frontend
+++ b/docker/Dockerfile.frontend
@@ -2,7 +2,7 @@
 
 ###############################################################################
 # Image with Node.js environment for frontend.
-FROM node:22.16-bookworm-slim AS frontend-env
+FROM node:22.19-bookworm-slim AS frontend-env
 
 # Set up certificates for any proxies that can get in the middle of curl/wget commands during the build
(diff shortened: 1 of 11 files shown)


sei-secheverria (changes requested):
Frontend compilation is failing (as seen in Docker CI task), which is preventing me from testing this. I think the .lock file is out of date.

sei-aderr:
Should be good now, older version of node we had in the docker version wasn't supported by the new versions of the new csv libraries

sei-secheverria (approved):
It looks ok, tested ok with simple csv file.

PR #919, opened 2026-09-09
Adjust dataset field input to work better with many inputs

### LINKED ISSUE #916: Improve display of large number of fields in card data
Follow up to #915 

Collaborator has dataset with 100+ fields, once they can import it through the other task they will want to view it in the card. This will be very difficult with the current structure of the card. Consider options for how this could be improved and implement one.

Options
- Tabular
- Show the same view as one now, but have a select for which is showing?
  - This could be nice, but then how would I handle wanting to view more than 1 at a time

sei-aderr:
Resolves #916


sei-secheverria (approved):
Looks good, merging it for now, we can create a separate PR if we want to change it.

